Issue #4, opened 2021-03-05
Seeing My Submissions


cronokirby:
As a Seller, I want to be able to see the items that I've put for sale, to get an overview of what I'm selling, and to make sure the description is correct

PR #15, opened 2021-03-06
Add a basic description form

### LINKED ISSUE #2: Make a form for entering item descriptions

### LINKED ISSUE #3: Describing Items
As a Seller, I can submit a textual description of an item, so that potential Buyers can see it

cronokirby:
Fixes #2
Fixes #3

# Changes

This just creates a second activity, where the user can enter a description for an item. After entering the description, the user can then see what they just entered on the main screen. With #4, you'll have a richer view, where you'll be able to see more than one item, but that's for a different story.

image
image

# Implementation

I used the [Data Binding](https://developer.android.com/topic/libraries/data-binding) feature for Android, which is a convenient way to tie data and the view. This also scales well to ViewModels, from what I've read so far.

There's also a few integration tests thrown in, so that can serve as a nice example for how to do those.


app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -1,11 +1,22 @@
...
+import androidx.databinding.DataBindingUtil
+import com.example.sharingang.databinding.ActivityMainBinding
 
 class MainActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_main)
+        val binding: ActivityMainBinding =
+            DataBindingUtil.setContentView(this, R.layout.activity_main)
+       binding.description = intent.getStringExtra("LAST_DESCRIPTION")
```
v1ruso, line 15: Here, how can you access intent without declaring it?

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -1,11 +1,22 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import androidx.databinding.DataBindingUtil
+import com.example.sharingang.databinding.ActivityMainBinding
 
 class MainActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_main)
+        val binding: ActivityMainBinding =
+            DataBindingUtil.setContentView(this, R.layout.activity_main)
```
v1ruso, line 14: Why is "binding" just a local variable, whereas in NewItemActivity it is an attribute?

app/src/main/java/com/example/sharingang/MainActivity.kt
v1ruso, line 15: How can you access intent without declaring it first?

app/src/main/java/com/example/sharingang/NewItemActivity.kt
```diff
@@ -0,0 +1,26 @@
...
+class NewItemActivity : AppCompatActivity() {
+    private lateinit var binding: ActivityNewItemBinding
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        binding = DataBindingUtil.setContentView(this, R.layout.activity_new_item)
+    }
+
+    fun createNewItem(view: View) {
+        val intent = Intent(baseContext, MainActivity::class.java)
```
v1ruso, line 19: What is "baseContext"?

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: So there's a [getIntent](https://developer.android.com/reference/android/app/Activity.html#getIntent()) method on Activity, which you have access to since you're inheriting from that class (indirectly).

In Kotlin, it's customary to use [properties](https://kotlinlang.org/docs/properties.html) instead.

The final bit of magic, is that Android's APIs, through the Kotlin shims wrapping them, have properties inserted instead of `getX` `setX` pairs.

So, instead of calling `this.getIntent()`, you can instead just write `this.intent`, or even `intent` as a shorthand.

app/src/main/java/com/example/sharingang/NewItemActivity.kt
```diff
@@ -0,0 +1,26 @@
...
+        intent.putExtra(
+            "LAST_DESCRIPTION",
+            binding.description
+        )
```
lumapools, line 23: I think it would be cleaner to declare "LAST DESCRIPTION" as a public static final String of NewItemActivity.

app/src/main/java/com/example/sharingang/NewItemActivity.kt
cronokirby: So you need some kind of `Context` class when starting an Intent, and `baseContext` (shorthand for `this.getBaseContext()`) is one way of getting at that. (I can't honestly claim to understand how contexts work that well, there's like 4 methods to get a context, and this was the one that happened to be used when I looked up how to start an Intent)

That being said, I think using `this` instead would also work, since Activities are also Contexts, and would probably be a lot clearer.

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: In `NewItemActivity`, it needs to be an attribute, since we need to reference it later in the class. We use the description when creating our Intent. Here, we don't use it anywhere else in the activity, so it's better to just keep things as a local variable to this method.

lumapools (changes requested):
Suggestion of changing declaration of "LAST DESCRIPTION" in NewItemActivity

app/src/main/java/com/example/sharingang/NewItemActivity.kt
```diff
@@ -0,0 +1,26 @@
...
+        startActivity(intent)
```
betrisey, line 21: NewItemActivity shouldn't start the MainActivity, instead it can call setResult() with the intent and then finish()

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -1,11 +1,22 @@
...
+       binding.description = intent.getStringExtra("LAST_DESCRIPTION")
+    }
+
+    fun createNewItem(view: View) {
+        val intent = Intent(this, NewItemActivity::class.java)
+        startActivity(intent)
```
betrisey, line 20: And here we should wait for the result of NewItemActivity, the best way seems to be like this: https://stackoverflow.com/a/66345734/8077618

app/src/main/java/com/example/sharingang/NewItemActivity.kt
cronokirby: Indeed.

I think it would be even better to use a standard intent extra key, like `Intent.EXTRA_TEXT`, having read a bit more:
https://developer.android.com/guide/components/intents-filters.

betrisey (changes requested):
After adding an item, the NewItemActivity is never terminated and a new MainActivity is created.

https://developer.android.com/training/basics/intents/result#register
https://stackoverflow.com/a/66345734/8077618

Everything else looks good to me! 👍

cronokirby:
> After adding an item, the NewItemActivity is never terminated and a new MainActivity is created.
> 
> https://developer.android.com/training/basics/intents/result#register
> https://stackoverflow.com/a/66345734/8077618
> 
> Everything else looks good to me! +1

That was indeed a good idea! Pretty easy to change, although now the integration test has to start from the main activity, which makes more sense anyways.

app/src/main/java/com/example/sharingang/NewItemActivity.kt
lumapools: Agreed, perfect. For some reason, I cannot find (or I may be blind) the declaration of EXTRA_TEXT, where is it located?

betrisey (approved):
Thanks

app/src/main/java/com/example/sharingang/MainActivity.kt
v1ruso: Perfect explanation, thanks!

app/src/main/java/com/example/sharingang/NewItemActivity.kt
v1ruso: Since I'm probably not the only one in the group unfamiliar with Kotlin, I think it might be less confusing to avoid using the shorthands notations (aka use this. instead of just the property), if you don't mind.

v1ruso (approved):

lumapools (approved):

PR #25, opened 2021-03-08
View multiple items at once

### LINKED ISSUE #11: Add a simple view after creating an item
This won't need any persistence

See #4

cronokirby:
Fixes #11 

This also adds some architectural patterns, like using a ViewModel for the logic.

This was made more complicated by figuring out the incantations necessary to get RecyclerView to work, but that's not all that important.


app/src/main/java/com/example/sharingang/items/ItemsAdapter.kt
```diff
@@ -0,0 +1,28 @@
...
+        set(value) {
+            field = value
+            notifyDataSetChanged()
+        }
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ItemViewHolder {
+        val layoutInflater = LayoutInflater.from(parent.context)
+        val view = layoutInflater.inflate(R.layout.item_view, parent, false) as TextView
+        return ItemViewHolder(view)
+    }
```
lumapools, line 20: How does does this method (`onCreateViewHolder`) work? I am not sure I understand its functionality

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -5,28 +5,30 @@ import android.content.Intent
...
+import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.databinding.DataBindingUtil
 import com.example.sharingang.databinding.ActivityMainBinding
 import com.example.sharingang.items.Item
 import com.example.sharingang.items.ItemsViewModel
 
 class MainActivity : AppCompatActivity() {
-    private lateinit var binding: ActivityMainBinding
+    private val viewModel: ItemsViewModel by viewModels()
```
lumapools, line 18: Could you please explain in a few words why the choice of ViewModels and how it works in our case?

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
```diff
@@ -0,0 +1,28 @@
+package com.example.sharingang.items
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+/**
+ * ItemsViewModel models the state of the fragment for viewing items
+ */
```
lumapools, line 9: In the comment, what does "fragment" mean?

betrisey (approved):

app/src/test/java/com/example/sharingang/LiveDataTestUtil.kt
```diff
@@ -0,0 +1,41 @@
...
+            throw TimeoutException("LiveData value was never set.")
+        }
+
+    } finally {
+        this.removeObserver(observer)
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    return data as T
+}
```
lumapools, line 41: I don't really understand how this class works or what its purpose is. What is it useful for? Could you please describe the big picture in 1-2 sentences?

app/src/main/java/com/example/sharingang/NewItemActivity.kt
```diff
@@ -17,7 +17,7 @@ class NewItemActivity : AppCompatActivity() {
     }
 
     fun createNewItem(view: View) {
-        val intent = Intent(this, MainActivity::class.java)
+        val intent = Intent()
```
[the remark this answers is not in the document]
cronokirby: Instead of explicitly declaring that we want to launch the MainActivity, we just say that we want to return a result from this activity, ending up at whatever activity called us.

In practice, this gives us the same result, since MainActivity is the only thing launching NewItemActivity for its result, but if another activity were to do the same, explicitly using MainActivity in the intent here wouldn't return to the right place, I think.

app/src/main/java/com/example/sharingang/items/ItemsAdapter.kt
cronokirby: So, for a list of different items, each of the items you display needs to create a view / widget on the screen. Because of this, you have `ItemViewHolder`, which encapsulates this widget. Basically, this section of code creates this widget dynamically, using the description in `item_view.xml`, which describes a single `TextView` widget.

You need to get a `LayoutInflater` from whatever surrounding context you have, in order to take the description of the layout for this widget in the xml file, and actually create a widget on the screen.

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: One advantage of a ViewModel is that the logic can be extracted from the activities / fragments, and tested independently. Furthemore, you can have a view model encapsulate logic such as accessing a database, which is going to be useful pretty soon. You can also share a viewmodel between fragments, which is convenient.

In our case, this isn't strictly necessary, but since we have a list of items, it's already a bit of a good idea to encapsulate that list into a view model, which just provides a method to add items, and a way to view the current items.

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
cronokirby: A fragment is the section of an activity used for displaying things. Concretely, this comment might be slightly misplaced, because there's no separate fragment from MainActivity right now, but the idea is that this state models that of the section of the activity responsible for displaying items.

app/src/test/java/com/example/sharingang/LiveDataTestUtil.kt
cronokirby: So, you use `LiveData` in your ViewModel instead of just a normal class, because you want the UI to be able to listen to updates / changes in the data.

The problem with testing this directly, is that now you have asynchronicity, because setting a new value isn't guaranteed to be immediately visible, because LiveData is designed to be used in the setting of an actual app.

This provides a `.getOrAwaitValue` method for the class `LiveData`, in order to block the current thread, waiting for the latest change. This exists only to make testing these classes possible, allowing you to turn `LiveData<T>` into `T` by blocking, and waiting for the latest result.

In all honesty, I don't understand how this snippet of code works, it was in the guide I read about Android testing:
https://developer.android.com/codelabs/advanced-android-kotlin-training-testing-basics#8

I think this is essentially the recommended way of testing `LiveData`, but it's not exactly pretty, and would probably be worthy of including into the standard Android extensions, but I digress.

app/src/main/java/com/example/sharingang/items/ItemsAdapter.kt
lumapools: Perfectly clear, thanks!

lumapools (approved):

PR #26, opened 2021-03-09
Add documentation about Firebase

### LINKED ISSUE #7: Research Firebase
See how it works exactly and if this is appropriate for the project

We'll we use real-time database or Cloud Firestore?

Part of user story #5

### LINKED ISSUE #8: Design database
See #5

betrisey:
Closes #7 
Closes #8


cronokirby (approved):

bafavre:
Do we already need to consider that maybe users want to "star" some items they're interested in? Like having a list of favorites. Would that change the model?

doc/firebase/firebase.md
```diff
@@ -0,0 +1,54 @@
...
+All queries can either
+- be run once to get the data
+- be "observed" and get notified in real time when a document matching the query changes (will be useful for the chat feature)
+
+## Data model
+We will have an `items` collection containing item documents with the attributs defined below.
+
+![diagram](img/model.png)
+
+In the future we will also have a `users` collection and each item will have the userID of its creator.
```
bafavre, line 37: Do we already need to consider that maybe users want to "star" some items they're interested in? Like having a list of favorites. Would that change the model?

betrisey: I wasn't sure either if we should already think about the fields for future features so I only added the fields for the main features.
But you're right, it's important to consider it now to avoid having to change the model later.

For the list of favorites, I think we can simply add a `favorites` field to the user which contains an array of item's IDs.
I've just found out that we'll be able to retrieve all the items in the list of favorite using a [whereIn clause](https://firebase.google.com/docs/firestore/query-data/queries?hl=en#in_not-in_and_array-contains-any). I'll add that to the doc :)

bafavre (approved):

PR #27, opened 2021-03-10
Add database using Firestore

### LINKED ISSUE #5: Seeing Other Submissions
As a Buyer, I want to see the items that other Sellers have created, to be able to potentially buy them

### LINKED ISSUE #9: Create classes that interact with database

betrisey:
To test it, you'll need to install [Firebase CLI](https://firebase.google.com/docs/cli?hl=en)
And then run this command from the root directory of the project:
`firebase emulators:start --only firestore`

Closes #5
Closes #9


app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
```diff
@@ -0,0 +1,85 @@
...
+
+class FirestoreItemRepository(useFirebaseEmulator: Boolean = false) : ItemRepository {
+    private val firestore = Firebase.firestore
+    private val collectionName = "items"
+
+    init {
+        if (useFirebaseEmulator) {
+            // 10.0.2.2 is the special IP address to connect to the 'localhost' of
+            // the host computer from an Android emulator.
+            firestore.useEmulator("10.0.2.2", 8080)
```
betrisey, line 23: If you are testing on a real phone, you might need to replace it with the IP of your computer

cronokirby:
I changed the code climate settings to allow coverage to decrease, try committing something else

betrisey:
> I changed the code climate settings to allow coverage to decrease, try committing something else

Okay thanks!

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+            .get()
+            .await()
+
+        return if (document != null) {
+            document.toObject(Item::class.java)
+        } else {
+            Log.d(TAG, "No Item with ID = $id")
+            null
+        }
+    }
```
cronokirby, line 58: IMO, separate logging from return value here. Plus, you can use Kotlin's neat nullable features :)

```suggestion
        if (document != null) {
          Log.d(TAG, "No Item with ID = $id")
        }
        return document?.toObject(Item::class.java)
    }
```

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+            .orderBy("createdAt", Query.Direction.DESCENDING)
+
+        val itemsLiveData = MutableLiveData<List<Item>>()
+
+        query.addSnapshotListener { value, error ->
+            if (error != null) {
+                Log.e(TAG, "Failed to get all items from Firebase.", error)
+            }
+
+            itemsLiveData.value = value!!.map { it.toObject(Item::class.java) }
```
cronokirby, line 80: I don't think `Log.e` exits the program, so if there's an error, and `value` is null, then you get a NullPointException because of the `value!!` bit later. I think you need to do an early return from the lambda in case of an error.

app/src/main/java/com/example/sharingang/items/Item.kt
```diff
@@ -1,6 +1,23 @@
 package com.example.sharingang.items
 
+import java.util.*
```
cronokirby, line 3: I think you just need `java.util.Date` here, so I'd restrict the imports

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
betrisey: Thanks, it's much more readable like that :)

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+            .document(id)
+            .get()
+            .await()
+
+        return if (document != null) {
+            document.toObject(Item::class.java)
+        } else {
+            Log.d(TAG, "No Item with ID = $id")
+            null
+        }
```
cronokirby, line 57: IMO, separate logging from return value here. Plus, you can use Kotlin's neat nullable features :)

```suggestion
        if (document != null) {
          Log.d(TAG, "No Item with ID = $id")
        }
        return document?.toObject(Item::class.java)
```

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+
+        val itemsLiveData = MutableLiveData<List<Item>>()
+
+        query.addSnapshotListener { value, error ->
+            if (error != null) {
+                Log.e(TAG, "Failed to get all items from Firebase.", error)
+            }
+
+            itemsLiveData.value = value!!.map { it.toObject(Item::class.java) }
+        }
```
cronokirby, line 81: I don't think `Log.e` exits the program, so if there's an error, and `value` is null, then you get a NullPointException because of the `value!!` bit later. I think you need to do an early return from the lambda in case of an error.

app/src/main/java/com/example/sharingang/items/Item.kt
```diff
@@ -1,6 +1,23 @@
...
+    /** URLs of the images */
+    val images: List<String> = listOf(),
+
+    /** Price in cents */
+    val price: Int = 0,
+
+    val createdAt: Date = Date(),
+
+    /** ID generated by the database */
+    var id: String? = null
```
cronokirby, line 22: I'm unsure that we want to allow `title` and `description` to default to `""`. I can't imagine a situation where an item without a title would be valid. Description, sure, but we probably should require it to be passed anyway.

For price `0` as a default does make sense, because we do have sharing of many items without a sale, so yeah.

app/src/main/java/com/example/sharingang/items/ItemRepository.kt
```diff
@@ -0,0 +1,30 @@
...
+     */
+    suspend fun addItem(item: Item): String?
+
+    /**
+     * Returns the item with corresponding id or null if it doesn't exist
+     */
+    suspend fun getItem(id: String): Item?
+
+    /**
+     * Returns a LiveData of all items. When the database is modified,
```
cronokirby, line 26: Nitpick: probably better to not use the term "LiveData", but rather describe what you're actually returning. The fact that it's LiveData is visible from the type signature.

Also, the secondary comment should be a couple lines down, like you do with `addItem`'s comment

app/src/main/java/com/example/sharingang/items/ItemRepositoryModule.kt
```diff
@@ -0,0 +1,19 @@
...
+import dagger.Binds
+import dagger.Module
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ViewModelComponent
+
+@Module
+@InstallIn(ViewModelComponent::class)
+/**
+ * Tells Hilt to use an instance of FirestoreItemRepository when we require an ItemRepository
+ */
```
cronokirby, line 12: Nitpick: I think the convention is to put doc comments *before* annotations, and we already did it that way in `ItemsViewModel`

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
```diff
@@ -0,0 +1,100 @@
...
+
+private const val TAG = "FirestoreItemRepository"
+
+@Singleton
+/**
+ * Implementation of ItemRepository using the Firestore database
+ *
+ * During development it requires running the Firebase emulator (see README.md)
+ */
+class FirestoreItemRepository @Inject constructor() :
```
cronokirby, line 24: Nitpick: I think the convention is to put doc comments *before* annotations, and we already did it that way in `ItemsViewModel`

app/src/test/java/com/example/sharingang/items/InMemoryItemRepositoryTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Test
+    fun inMemoryItemRepositoryStartsEmpty() {
+        val repo: ItemRepository = InMemoryItemRepository()
+        runBlocking {
+            assert(repo.getAllItems().isEmpty())
+        }
+    }
+
+    @Test
+    fun inMemoryItemRepositoryCanRetrieveStoredData() {
```
cronokirby, line 22: it might be a good idea to split this test up?

The description is a bit vague, and there's quite a few different assertions.

cronokirby (changes requested):
Great job overall!

I think the repository abstraction is great, and integrating Hilt for dependency injection to use an in memory repository for tests was done nicely.

I just have a few minor code suggestions, some formatting nitpicks, and perhaps the test for the in memory repository could be split up a bit

Anyways, cool beans! 8^)

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
betrisey: Thanks, it's much more readable :)

app/src/main/java/com/example/sharingang/items/Item.kt
betrisey: You're right it doesn't make sense to have an object without a title.
I had to do this because Firebase requires having a no-argument constructor to deserialize it with `.toObject(Item::class.java)`.
Can we keep it like this for now and maybe add manual deserialization when we have more complicated fields (e.g. geo coordinates)?

cronokirby: Ah, yeah that makes sense. Sure, it's fine for now, just a minor point.

app/src/test/java/com/example/sharingang/items/InMemoryItemRepositoryTest.kt
betrisey: Yes, I'll split it.

cronokirby (approved):
Nice!

betrisey:
> Nice!

Thanks for the review!

firestore.rules
```diff
@@ -0,0 +1,9 @@
+rules_version = '2';
+service cloud.firestore {
+  match /databases/{database}/documents {
+    match /{document=**} {
+      allow read, write: if
+          request.time < timestamp.date(2021, 4, 4);
```
bafavre, line 6: Why is there this time rule here? I'm not sure I understand why we need to be less than 4.4.2021

betrisey: It's a generated automatically by Firebase, it allows everything for a month.
We'll have to change it and specify real permissions when we'll have user authentication.

bafavre (approved):

PR #28, opened 2021-03-10
Add setting price view

lumapools:
Added a view to set the price. When we add the price, we are then directed to another activity that shows the price we entered. Clicking on the "Back" arrow will take us back to the Main Activity.

Screenshot from 2021-03-10 16-31-56

Screenshot from 2021-03-10 16-32-08

When a user clicks on OK while the field is blank, the price is automatically set to 0.00 USD.

Screenshot from 2021-03-10 16-32-34

Screenshot from 2021-03-10 16-32-45

Interaction tests have been added as well.


app/src/main/java/com/example/sharingang/SetPriceActivity.kt
```diff
@@ -0,0 +1,32 @@
...
+import android.content.Intent
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+import android.view.View
+import android.widget.EditText
+
+class SetPriceActivity : AppCompatActivity() {
+    companion object {
+        const val EXTRA_PRICE = "com.example.sharingang.EXTRA_PRICE"
+        var CHOSEN_CURRENCY = "USD"
```
cronokirby, line 12: This isn't ever modified so it really shouldn't be a var. I think `const val` would be the most appropriate here.

app/src/main/java/com/example/sharingang/SetPriceActivity.kt
```diff
@@ -0,0 +1,32 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_set_price)
+    }
+
+    fun setPriceSummary(view: View) {
+        val intent = Intent(this, SetPriceSummaryActivity::class.java)
+        val editTextPrice : EditText = findViewById(R.id.editTextSetPrice)
+        val price =
+            if(editTextPrice.text.isNullOrBlank()) defaultPrice // if no price has been entered
```
cronokirby, line 26: I think you've forgotten to format this file, because there would likely be a space between the if and the parens here, and there would be no space between `editTextPrice` and `:`

app/src/main/java/com/example/sharingang/SetPriceActivity.kt
```diff
@@ -0,0 +1,32 @@
...
+            else editTextPrice.text.toString().toDouble()
+
+        intent.putExtra(EXTRA_PRICE, "The price you set is: $price $CHOSEN_CURRENCY.")
```
cronokirby, line 29: I don't think it's great that you're choosing how to display the information by sending a formatted string through the intent.

I think it would make more sense to send out a double representing the price instead. Perhaps combine one extra for the amount, and another for the currency, if you want to support multiple currencies. Regardless, sending out a pre-formatted string seems like a bad idea, because the activity displaying the price should be in control of how it does that.

app/src/main/res/layout/activity_set_price_summary.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
cronokirby, line 2: Consider using just a `LinearLayout` here instead, or a `FrameLayout`, I think it's called, which contains just a single component. Simpler since you don't need `ConstraintLayout` yet, but I don't think it changes much.

cronokirby (changes requested):
One change I think should be done is to avoid sending a formatted string across the intent, and displaying that naively. Instead, an amount (and perhaps currency) should be sent, and then that should be used instead, so that you don't couple the information generated by one activity, and how another chooses to display it.

cronokirby:
Also, you'll need to either rebase your changes against the main branch, or merge the main branch into this branch, to resolve the conflict you have currently.

LMK if you'd like help with that :)

app/src/main/java/com/example/sharingang/SetPriceActivity.kt
lumapools: Okay! So instead of :
`intent.putExtra(EXTRA_PRICE, "The price you set is: $price $CHOSEN_CURRENCY.")`
I can have something like
`intent.putExtra(EXTRA_PRICE, price)`?
or even
`intent.putExtra(EXTRA_PRICE, Pair(price, CHOSEN_CURRENCY))` ?

cronokirby: Yeah, that's what I'm getting at!

cronokirby (approved):
Nice!

bafavre (approved):

PR #30, opened 2021-03-11
Edit items

### LINKED ISSUE #13: Add a view to edit the Item
Seeing the old value before edit and the updated one after

See #6

bafavre:
Fixes #13 

# Changes

Replaced all the activities by fragments and only keeping one main activity. Added a Button to the item layout used in the RecycledView to navigate to a new fragment to be able to edit an item.

image
image
image

# Implementation

I followed the [Fragment and Navigation](https://developer.android.com/codelabs/kotlin-android-training-create-and-add-fragment) codelabs to adapt our existing app and use fragments and navigation instead of multiple activities.
I also followed the [RecyclerView](https://developer.android.com/codelabs/kotlin-android-training-recyclerview-fundamentals?index=..%2F..android-kotlin-fundamentals) codelabs to add a new button to the RecyclerView and interact with it to reach the new fragment and be able to edit the item.


betrisey (commented):
Great work, thanks!

betrisey (approved):

app/src/main/java/com/example/sharingang/EditItemFragment.kt
```diff
@@ -0,0 +1,40 @@
...
+    private lateinit var item: Item
+
+    override fun onCreateView(
+        inflater: LayoutInflater, container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        val binding: FragmentEditItemBinding =
+            DataBindingUtil.inflate(inflater, R.layout.fragment_edit_item, container, false)
+
+        val args = EditItemFragmentArgs.fromBundle(requireArguments())
```
cronokirby, line 28: Smart.

app/src/main/java/com/example/sharingang/items/ItemListener.kt
```diff
@@ -0,0 +1,6 @@
+package com.example.sharingang.items
+
+// Replace the description with its ID once the database is set
+class ItemListener(val clickListener: (item: Item) -> Unit) {
```
cronokirby, line 4: Question: does this have to be outside of the view holder?

Otherwise, it might be a bit cleanser to move it inside; less surface area

cronokirby (approved):
Very very nice!

Seems like it was quite the royal pain in the ass to make the list mutable LOL. Definitely took one for the team there.

One question that might be cool if it doesn't take long is to try and move the Item Listener class inside of the view holder, that way there's one less class to deal with.

app/src/main/java/com/example/sharingang/items/ItemListener.kt
bafavre: No it doesn't, actually in the codelabs they put everything inside the adapter: the ViewHolder, Listener and DiffCallback.

I thought of doing this first and then noticed that you put the ViewHolder in its own file so I thought I would also create a different one for every class.

Should I move everything inside the adpater as they did or just the Listener inside the ViewHolder? What do you think?

PR #45, opened 2021-03-14
Add editing for title and description

cronokirby:
Instead of just having the description, I added an extra title field, and made the description field a bit larger:

I also noticed that data binding wasn't used to its fullest potential everywhere, so I took the liberty of making that change as well.


app/src/main/res/layout/fragment_edit_item.xml
```diff
@@ -1,5 +1,17 @@
...
+    <data>
+
+        <variable
+            name="title"
+            type="String" />
+
+        <variable
+            name="description"
+            type="String" />
+    </data>
```
bafavre, line 13: Good call!

bafavre (approved):
All good, well done! :)

v1ruso (approved):

PR #48, opened 2021-03-15
Add price to items

lumapools:
Prices can now be added to items when creating or editing them. If the user does not enter a price, it is set to 0.0 by default. When a user sets a price and then edits the item, the previous price is displayed when editing it. 

Webp net-resizeimage  Webp net-resizeimage (1)

### DIFF
--- a/app/src/main/java/com/example/sharingang/NewItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/NewItemFragment.kt
@@ -29,7 +29,7 @@ class NewItemFragment : Fragment() {
                 Item(
                     description = binding.description ?: "",
                     title = binding.title ?: "",
-                    price = binding.price
+                    price = binding.price?.toDoubleOrNull() ?: 0.0
 
                 )
             )
--- a/app/src/main/java/com/example/sharingang/NewItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/NewItemFragment.kt
@@ -28,7 +28,9 @@ class NewItemFragment : Fragment() {
             viewModel.addItem(
                 Item(
                     description = binding.description ?: "",
-                    title = binding.title ?: ""
+                    title = binding.title ?: "",
+                    price = binding.price
+
                 )
             )
             view.findNavController().navigate(R.id.action_newItemFragment_to_itemsListFragment)
--- a/app/src/main/java/com/example/sharingang/NewItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/NewItemFragment.kt
@@ -27,10 +27,9 @@ class NewItemFragment : Fragment() {
         binding.createItemButton.setOnClickListener { view: View ->
             viewModel.addItem(
                 Item(
+                    price = binding.price?.toDoubleOrNull() ?: 0.0,
                     description = binding.description ?: "",
-                    title = binding.title ?: "",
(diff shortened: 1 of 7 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -8,4 +8,6 @@
     <string name="edit">Edit</string>
     <string name="edit_item">Edit Item</string>
     <string name="title">Title</string>
+    <string name="default_price">Price (default: 0.0)</string>
+    <string name="price">Price</string>
```
v1ruso, line 12: Where is this string used?

app/src/main/java/com/example/sharingang/NewItemFragment.kt
```diff
@@ -28,7 +28,9 @@ class NewItemFragment : Fragment() {
             viewModel.addItem(
                 Item(
                     description = binding.description ?: "",
-                    title = binding.title ?: ""
+                    title = binding.title ?: "",
+                    price = binding.price?.toDoubleOrNull() ?: 0.0
+
                 )
```
v1ruso, line 34: Same code in both EditItemFragment and NewItemFragment

v1ruso:
Well done:)

app/src/main/res/values/strings.xml
lumapools: When creating a new item, in NewItemFragment I think

lumapools: "Price" is used when we are editing items and we completely clear the field where the price goes. In this case, the hint "Price" gets displayed in the editText box.

app/src/main/java/com/example/sharingang/items/Item.kt
```diff
@@ -18,8 +18,7 @@ data class Item(
     /** URLs of the images */
     val images: List<String> = listOf(),
 
-    /** Price in cents */
-    val price: Int = 0,
+    val price: Double = 0.0,
```
bafavre, line 21: I think the price being an int and using cents was intentional, I don't know if we should change that or not (some currencies don't have the notion of "cents" or commas used in their money). Maybe check with @betrisey why it was an int before.

bafavre (commented):
Well done! :)

app/src/main/java/com/example/sharingang/items/Item.kt
lumapools: We discussed it with @betrisey and agreed on the fact that a `double` should be fine. The reason for using an `int` (as cents) was to avoid problems with floating point arithmetic. A price will only be set or read, therefore no floating point arithmetic issues should occur.

app/src/main/java/com/example/sharingang/NewItemFragment.kt
lumapools: It is indeed the same code, but refactoring this would be quite bothersome, and would maybe lead to unintended results. Both classes use this piece of code, however the `binding` in `NewItemFragment` is different from `binding` in `EditItemFragment`. In one case, it is a `FragmentNewItemBinding` and in the second case it is a `FragmentEditItemBinding`. I was thinking about making a common method that takes a `Binding` as argument and returns the Item based on this argument, but it is quite difficult as Bindings classes are automatically generated and there is no common superclass.

v1ruso (approved):

bafavre (approved):

PR #49, opened 2021-03-15
Map display

### LINKED ISSUE #35: Display a map

v1ruso:
Fix #35

Once clicking on the "go to map" button in the main activity, the current activity switches to the map activity.

If the permission for location hasn't been granted yet, a pop-up appears, requesting for it:
request_permission

Once the permission for location has been granted, the current location of the device is displayed both textually and on the map. The location is updated automatically every 2.5 to 5 seconds.
map_display


app/src/androidTest/java/com/example/sharingang/MapActivityTest.kt
```diff
@@ -27,29 +25,28 @@ class MapActivityTest {
...
+    fun grantingPermissionDisplaysLocation() {
+        grantPermission()
         val text = onView(withId(R.id.location_display))
         text.check(matches(withText("")))
-
-        button.perform(click())
-        grantPermission()
-        button.perform(click())
+        Thread.sleep(10000)
+        text.check(matches(withText(containsString("Your location"))))
```
betrisey, line 50: Does anyone know a way to wait for the text to change without sleeping?
I've found something using IdleResources but it requires changes in the class specifically for the test which it not very good.

app/src/main/res/layout/activity_map.xml
```diff
@@ -8,25 +8,23 @@
...
 
-        <TextView
-            android:id="@+id/location_display"
-            android:layout_width="183dp"
-            android:layout_height="158dp"
-            app:layout_constraintBottom_toBottomOf="parent"
+        <com.google.android.gms.maps.MapView
+            android:id="@+id/mapView"
+            android:layout_width="407dp"
+            android:layout_height="391dp"
```
betrisey, line 23: It would be better not to hard-code the dimensions, maybe you can use match_parent for the width.
For the height, I'm not sure how. Maybe the height of the TextView could be wrap_content and the map use all the space left.

app/src/main/res/layout/activity_map.xml
```diff
@@ -8,25 +8,23 @@
...
             app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintHorizontal_bias="0.478"
```
betrisey, line 25: Not sure if this is intended

betrisey (changes requested):
Nice :)

app/src/main/res/layout/activity_map.xml
v1ruso: You're right I will fix that, thanks!

app/src/main/res/layout/activity_map.xml
v1ruso: Probably not, I will verify.

app/src/main/java/com/example/sharingang/MapActivity.kt
```diff
@@ -1,62 +1,156 @@
...
-            }
+    private fun startLocationUpdates() {
+        if (!permissionGranted) {
+            requestLocationPermission()
+        } else {
+            fusedLocationClient.requestLocationUpdates(
+                LocationRequest.create().apply {
+                    interval = 5000
+                    fastestInterval = 2500
+                    priority = LocationRequest.PRIORITY_HIGH_ACCURACY
```
lumapools, line 77: What is the difference between `interval` and `fastestInterval`? How do we choose which interval we are going to be using for updating the location?

betrisey (approved):

lumapools (approved):
Good! Nice work!

PR #51, opened 2021-03-16
Improve database integration and cleanup

### LINKED ISSUE #44: Improve database integration

betrisey:
I've check with @bafavre and improved a few things:
- `ItemsViewModel.updateItem` now only takes one argument, the updated Item which contains the `id`
- The `id` of the `Item` data class is now a `val`.
- Use copy method to create an updated Item (e.g. `item.copy(title = "new title")`). This way, the other fields are automatically copied.
- Formatting, typos, etc...

Closes #44


bafavre (approved):
All good! :)

app/src/androidTest/java/com/example/sharingang/MapActivityTest.kt
```diff
@@ -49,4 +49,4 @@ class MapActivityTest {
         Thread.sleep(10000)
         text.check(matches(withText(containsString("Your location"))))
     }
-}
\ No newline at end of file
+}
```
v1ruso, line 52: Very useful x)

app/src/main/java/com/example/sharingang/items/Item.kt
```diff
@@ -20,9 +21,10 @@ data class Item(
...
     val price: Double = 0.0,
 
-    val createdAt: Date = Date(),
+    @ServerTimestamp
+    val createdAt: Date? = null,
 
     /** ID generated by the database */
     @DocumentId
-    var id: String? = null
+    val id: String? = null
```
v1ruso, line 29: "val" here would mean that id can never be reassigned no?

Superdoude1999: I believe so but why would you need to change the id anyways? Plus that way you can use ids to compare items (as seen below) since you know they won't change

Superdoude1999 (approved):
I forgot to click the approve button last time sorry...

app/src/androidTest/java/com/example/sharingang/MapActivityTest.kt
betrisey: It was added when I reformatted, I think this is good to have new lines at the end of files
https://thoughtbot.com/blog/no-newline-at-end-of-file

app/src/main/java/com/example/sharingang/items/Item.kt
betrisey: Yes, at the beginning we used var because we had to set the id after the item was added to the database.
But I discovered a better way, we can create a copy of the item and only change the id like that: `item.copy(id = ...)`

app/src/main/java/com/example/sharingang/MapActivity.kt
```diff
@@ -153,4 +152,4 @@ class MapActivity : AppCompatActivity(), OnMapReadyCallback {
         super.onLowMemory()
         binding.mapView.onLowMemory()
     }
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
betrisey: How GitHub is happy haha
image

app/src/main/java/com/example/sharingang/items/FirestoreItemRepository.kt
```diff
@@ -101,7 +101,7 @@ class FirestoreItemRepository @Inject constructor() :
         requireNotNull(item.id)
 
         return try {
-            firestore.collection(collectionName).document(item.id!!)
+            firestore.collection(collectionName).document(item.id)
```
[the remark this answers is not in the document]
betrisey: When a value is nullable (for exemple String?), we cannot use it directly with something that accepts a non-nullable type. We can either check if it is not null then use it, or if we are really sure it is not null, we can use `!!` to cast to non-nullable (but it throws an exception if it contains a null value).
Here because we have requireNotNull(item.id), Kotlin understands that the value has been checked so the !! is not required.

PR #52, opened 2021-03-16
Add a detailed view for items

cronokirby:
Now you can click on an item in the list to see a detailed view, including the description.


betrisey (approved):
Great!

app/src/androidTest/java/com/example/sharingang/DetailedItemFragmentTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.test.espresso.assertion.ViewAssertions
+import androidx.test.espresso.matcher.ViewMatchers
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
```
v1ruso, line 14: For this class, I think it might be cleaner to import Espresso.method, instead of calling that method with Espresso.method in the tests

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
```diff
@@ -21,6 +21,14 @@ class ItemsViewModel @Inject constructor(
     val navigateToEditItem: LiveData<Item?>
         get() = _navigateToEditItem
 
+    private val _focusedItem = MutableLiveData<Item?>()
```
v1ruso, line 24: Just out of curiosity, what does it mean to have an "_" before some variable names? For mutable variables?

Superdoude1999 (approved):
Looks good!

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
bafavre: I think it's just a notation for private variable.

cronokirby:
If CI fails now I will explode

PR #53, opened 2021-03-17
Add image description

bafavre:
Everything is a bit messy, I would gladly accept some recommendations concerning the way to handle the permissions, to check if it's a storage permission or (for later) if it's a camera permission. I put it in the activity but if I don't know which permission was accepted/rejected inside the registerForActivityResult.

Also since we do not upload on a server yet nor have a local cache, I just temporarily store the Uri of the picture, which will change in the future when we'll have these features obviously.

Also, Code Climate doesn't seem to like all the duplicate code, I'll have to refactor.

This is the result :


app/src/main/java/com/example/sharingang/EditItemFragment.kt
```diff
@@ -15,8 +20,19 @@ import com.example.sharingang.items.ItemsViewModel
 class EditItemFragment : Fragment() {
 
     private val viewModel: ItemsViewModel by activityViewModels()
+
+    private lateinit var contextActivity: MainActivity
     private lateinit var existingItem: Item
 
+    private var imageUri = MutableLiveData<Uri>()
```
cronokirby, line 27: I don't think this should be in the Fragment directly, because I don't think it gets persisted when for example, the screen rotates, restarting the fragment.

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -1,7 +1,49 @@
...
+    var permissionGranted: Boolean = false
+
+    private val requestPermissionLauncher =
+        registerForActivityResult(RequestPermission()) { isGranted: Boolean ->
+            permissionGranted = if (isGranted) {
+                true
+            } else {
+                Toast.makeText(
+                    this,
+                    "Storage permission successfully denied. Feature is disabled.",
```
cronokirby, line 22: This doesn't really work out if we're trying to make this as a general abstraction for requesting permissions right, since we're coding in the message for storage permissions.

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -1,7 +1,49 @@
...
+                    Toast.LENGTH_LONG
+                ).show()
+                false
```
cronokirby, line 25: You could probably write this segment as like:

```kotlin
if (!isGranted) {
  doSomething()
}
isGranted
```

app/src/main/java/com/example/sharingang/NewItemFragment.kt
```diff
@@ -15,6 +20,16 @@ import com.example.sharingang.items.ItemsViewModel
 class NewItemFragment : Fragment() {
 
     private val viewModel: ItemsViewModel by activityViewModels()
+    private lateinit var contextActivity: MainActivity
```
cronokirby, line 23: This should probably be a more restricted interface.

cronokirby (commented):
I have some initial comments, but I'm taking this as a more of a draft PR, so we should probably comb this over more finely later.

As far as reducing the redundancy, there's like, two ways to go about it.

- Creating a file / class to contain common permission functionality for both fragments
- Merging the two fragments

The first option is probably the most straightforward, the second is probably desirable in the long term.

app/src/main/java/com/example/sharingang/EditItemFragment.kt
```diff
@@ -15,8 +20,19 @@ import com.example.sharingang.items.ItemsViewModel
 class EditItemFragment : Fragment() {
 
     private val viewModel: ItemsViewModel by activityViewModels()
+
+    private lateinit var contextActivity: MainActivity
```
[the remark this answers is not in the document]
bafavre: I did that because the functions that I used were in the MainActivity. I don't know how to put these functions elsewhere than in the MainActivity, because most of the functions I call need to have an activity.

app/src/main/java/com/example/sharingang/EditItemFragment.kt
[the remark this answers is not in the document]
bafavre: True, I'll simply use a binding for the uri.

app/src/main/java/com/example/sharingang/MainActivity.kt
bafavre: You're right, I'll change that.

bafavre:
Selecting an image only works the first time a fragment is used and then not anymore.
For example I can create a new item, add an image, it displays correctly, change the image and it still displays correctly.
However if I were to create the item and then try to create a new one, selecting an image from the gallery doesn't display it.
I need to investigate that.

betrisey (approved):
Good job!

Superdoude1999 (approved):
Looks good

PR #56, opened 2021-03-18
Remove boolean to check for permission

v1ruso:
Nothing really new here, just the way to check for permission is different. I removed the boolean, as mentioned by @bafavre.


Superdoude1999 (approved):
Nice

betrisey (approved):

PR #72, opened 2021-03-20
Add location to items

### LINKED ISSUE #62: Add a location to items.

v1ruso:
Fix #62 

In the create new item fragment, I added two text fields and a button. The two text fields should contain the latitude and longitude of the user.
before_fetch

The user can then either write the location directly, or get its current location using the "get location" button.
after_fetch

In Edit fragment, the user can edit textually the location of the user. I decided to have separate features between Edit and NewItem fragments to make them a bit different.

I also changed MapActivity into a fragment.

Although I know user usually don't know their location coordinates, I thought it was easier to make it this way first. The goal would be then to implement a feature where they can write a postal address, which would then be transformed internally into coordinates.


app/src/main/java/com/example/sharingang/NewItemFragment.kt
```diff
@@ -1,43 +1,95 @@
...
+            LocationRequest.PRIORITY_HIGH_ACCURACY,
+            cancellationTokenSource.token
+        )
+        getLocationTask.addOnCompleteListener {
+            if (it.isSuccessful && it.result != null) {
+                binding.latitude = it.result.latitude.toString()
+                binding.longitude = it.result.longitude.toString()
+            }
+        }
+    }
```
cronokirby, line 94: I think that maybe this could be extracted further, given the duplication with the other fragment?

app/src/main/java/com/example/sharingang/utils/PermissionRequest.kt
```diff
@@ -0,0 +1,54 @@
...
+            callback()
+        }
+    }
+}
+
+fun doOrGetPermission(
+    context: Context?,
+    fragment: Fragment,
+    permission: String,
+    callback: () -> Unit,
```
cronokirby, line 26: Maybe instead of just having a callback with no arguments, you could instead get the fine location inside of this function, and then have the callback take the location.

Well, maybe not inside of `doorGetPermission`, but you could make another wrapper specifically for the location, right.

cronokirby: If you had a callback that took in the location, then that would simplify the duplication between both `updateLocation` methods, since you would just have the fragment specific code consuming that location.

cronokirby:
I haven't looked at everything in detail, but I think there's a few little things we could do to abstract over permissions a tad more.

Otherwise, good work so far!

app/src/main/java/com/example/sharingang/NewItemFragment.kt
v1ruso: You're right, I will see what I can do with this

app/src/main/java/com/example/sharingang/utils/PermissionRequest.kt
v1ruso: Ok I see, it makes sense!

v1ruso: Actually, I wanted this function to be as general as possible, to be able to use it for all the methods that need certain permissions. For example, MapFragment needs to continuously get the location, whereas NewItem and EditItem only need it once.

cronokirby: Yeah, my point was to make another wrapper around this function which would take a more specific callback.

app/src/androidTest/java/com/example/sharingang/EditItemFragmentTest.kt
```diff
@@ -27,12 +27,12 @@ class EditItemFragmentTest {
     private val editedItem = "Edited"
 
     @Test
-    fun canEditCategoryOfItem(){
+    fun canEditCategoryOfItem() {
```
cronokirby, line 30:
```suggestion
    fun canEditItemCategory() {
```

cronokirby: Seems a bit more natural sounding this way

app/src/androidTest/java/com/example/sharingang/MainActivityTest.kt
```diff
@@ -22,8 +24,12 @@ class MainActivityTest {
     @get:Rule(order = 1)
     val activityRule = ActivityScenarioRule(MainActivity::class.java)
 
+    @get:Rule
+    val grantPermissionRule: GrantPermissionRule =
+        GrantPermissionRule.grant(Manifest.permission.ACCESS_FINE_LOCATION)
+
     @Test
-    fun clickingOnButtonStartsMapActivity(){
+    fun clickingOnButtonStartsMapFragment(){
```
cronokirby, line 32:
```suggestion
    fun clickingOnButtonOpensMap() {
```
I think speaking of things in terms of how the User sees them instead of how the Android dev sees them makes sense.

Note: this file was likely not formatted either

app/src/androidTest/java/com/example/sharingang/MapFragmentTest.kt
```diff
@@ -1,24 +1,38 @@
...
+    val grantPermissionRule: GrantPermissionRule =
+        GrantPermissionRule.grant(Manifest.permission.ACCESS_FINE_LOCATION)
 
     @Test
-    fun grantingPermissionDisplaysLocation() {
+    fun openingFragmentDisplaysLocation() {
+        onView(withId(R.id.go_to_map)).perform(click())
         val text = onView(withId(R.id.location_display))
         text.check(matches(withText("")))
         Thread.sleep(10000)
```
cronokirby, line 38: Do we really have to wait this long? Is there not a better way here?

app/src/androidTest/java/com/example/sharingang/NewItemFragmentTest.kt
```diff
@@ -49,4 +56,29 @@ class NewItemFragmentTest {
...
         onView(withText(secondItem)).check(matches(isDisplayed()))
     }
+
+    @Test
+    fun clickingOnGetLocationDisplaysLocation() {
+        onView(withId(R.id.newItemButton)).perform(click())
+        val button = onView(withId(R.id.new_item_get_location))
+        button.check(matches(withText("Get Location")))
+        button.perform(click())
+        Thread.sleep(5000)
```
cronokirby, line 66: Probably no way to avoid the sleep here, but it would be nice, if it's easy to do.

app/src/androidTest/java/com/example/sharingang/NewItemFragmentTest.kt
```diff
@@ -49,4 +56,29 @@ class NewItemFragmentTest {
...
+        onView(withId(R.id.write_latitude)).check(matches(not(withText(""))))
+        onView(withId(R.id.write_longitude)).check(matches(not(withText(""))))
+    }
+
+    @Test
+    fun aLocationCanBeWritten() {
```
cronokirby, line 72: déjà vu 8^)

app/src/main/AndroidManifest.xml
```diff
@@ -16,9 +16,8 @@
             android:value="@string/google_api_key" />
         <meta-data
             android:name="com.google.android.gms.version"
-            android:value="@integer/google_play_services_version"/>
-        <activity android:name=".MapActivity"
-            android:parentActivityName=".MainActivity"/>
+            android:value="@integer/google_play_services_version" />
+
```
cronokirby, line 20: Is this file formatted? the blank space looks odd to me

app/src/main/java/com/example/sharingang/EditItemFragment.kt
```diff
@@ -17,34 +17,45 @@ class EditItemFragment : Fragment() {
     private val viewModel: ItemsViewModel by activityViewModels()
     private lateinit var existingItem: Item
 
+    private lateinit var binding: FragmentEditItemBinding
+
     override fun onCreateView(
-            inflater: LayoutInflater, container: ViewGroup?,
-            savedInstanceState: Bundle?
+        inflater: LayoutInflater, container: ViewGroup?,
+        savedInstanceState: Bundle?
```
cronokirby, line 24: This formatting looks odd to me, has this file been formatted? I'd expect to see the items on multiple lines, if the last item has been moved to a new line

app/src/main/java/com/example/sharingang/EditItemFragment.kt
```diff
@@ -17,34 +17,45 @@ class EditItemFragment : Fragment() {
...
-                            price = binding.price?.toDoubleOrNull() ?: 0.0,
-                            category = binding.categorySpinner.selectedItemPosition,
-                            categoryString = resources.getStringArray(R.array.categories)[binding.categorySpinner.selectedItemPosition]
-                    )
+                existingItem.copy(
+                    title = binding.title ?: "",
+                    description = binding.description ?: "",
+                    price = binding.price?.toDoubleOrNull() ?: 0.0,
+                    category = binding.categorySpinner.selectedItemPosition,
+                    categoryString = resources.getStringArray(R.array.categories)[binding.categorySpinner.selectedItemPosition],
```
cronokirby, line 42: Probably best to move this to a value, to avoid making a mega large line?

app/src/main/java/com/example/sharingang/EditItemFragment.kt
```diff
@@ -17,34 +17,45 @@ class EditItemFragment : Fragment() {
...
     }
+
+    private fun setupBinding() {
+        binding.title = existingItem.title
+        binding.description = existingItem.description
+        binding.price = existingItem.price.toString().format("%.2f")
+        binding.categorySpinner.setSelection(existingItem.category)
+        binding.latitude = existingItem.latitude.toString()
+        binding.longitude = existingItem.longitude.toString()
+
```
cronokirby, line 59:
```suggestion
```

Seems like odd formatting to me

app/src/main/java/com/example/sharingang/ItemsListFragment.kt
```diff
@@ -24,46 +20,50 @@ class ItemsListFragment : Fragment() {
...
                 viewModel.onViewItemNavigated()
             }
         })
     }
 
     override fun onCreateView(
-            inflater: LayoutInflater, container: ViewGroup?,
-            savedInstanceState: Bundle?
+        inflater: LayoutInflater, container: ViewGroup?,
+        savedInstanceState: Bundle?
```
cronokirby, line 40: Formatting looks off here, and otherwhere in the file

app/src/main/java/com/example/sharingang/utils/PermissionRequest.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.google.android.gms.tasks.CancellationTokenSource
+
+fun requestPermissionLauncher(
+    fragment: Fragment,
+    callback: () -> Unit
+): ActivityResultLauncher<String> {
+    return fragment.registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted: Boolean ->
+        if (isGranted) {
+            callback()
+        }
```
cronokirby, line 23:
```suggestion
    return fragment.registerForActivityResult(ActivityResultContracts.RequestPermission()) {
        if (it) {
            callback()
        }
```

I think this works?

app/src/main/java/com/example/sharingang/utils/PermissionRequest.kt
```diff
@@ -0,0 +1,76 @@
...
+fun requestPermission(
+    context: Context?,
+    fragment: Fragment,
+    permission: String,
+    resultLauncher: ActivityResultLauncher<String>
+) {
+    if (fragment.shouldShowRequestPermissionRationale(permission)) {
+        Toast.makeText(
+            context,
+            String.format("%s permission is required", permission),
```
cronokirby, line 53:
```suggestion
            "$permission permission is required"
```
I think this works

cronokirby (changes requested):
Great job overall!


Just a few minor nitpicks, and some questions about how things work.

cronokirby:
Also uh, we should try and get CI to pass huh 8^)

app/src/androidTest/java/com/example/sharingang/MapFragmentTest.kt
v1ruso: I probably can reduce it to 5 seconds, but I wasn't able to find a way to wait for the location result...

app/src/androidTest/java/com/example/sharingang/NewItemFragmentTest.kt
v1ruso: No clue how to do it unfortunately.

app/src/main/AndroidManifest.xml
v1ruso: Yes it is formatted

app/src/main/java/com/example/sharingang/EditItemFragment.kt
v1ruso: It is formatter as well

app/src/main/java/com/example/sharingang/EditItemFragment.kt
v1ruso: We should ask @Superdoude1999, this is not my code

app/src/main/java/com/example/sharingang/EditItemFragment.kt
v1ruso: The file is already formatted

app/src/main/java/com/example/sharingang/ItemsListFragment.kt
v1ruso: Already formatted apparently

app/src/main/java/com/example/sharingang/utils/PermissionRequest.kt
v1ruso: I just thought "isGranted" would be more understandable as to what this method does

app/src/main/java/com/example/sharingang/utils/PermissionRequest.kt
v1ruso: Yep thanks

app/src/main/java/com/example/sharingang/utils/PermissionRequest.kt
cronokirby: fair enough

cronokirby (approved):
Cool beans!

app/src/main/java/com/example/sharingang/MapFragment.kt
```diff
@@ -61,25 +52,20 @@ class MapActivity : AppCompatActivity(), OnMapReadyCallback {
         }
         binding.mapView.onCreate(savedInstanceState)
         binding.mapView.getMapAsync(this)
+        return binding.root
     }
 
-
+    @SuppressLint("MissingPermission")
```
bafavre, line 58: I don't quite remember, but didn't you say that there was a way to get rid of this? I think @betrisey mentionned it last week.

bafavre (approved):
Well done!

app/src/main/java/com/example/sharingang/MapFragment.kt
betrisey: What I mentioned was for this lint error that was a false positive and will be fixed with the next release of the AndroidX activity library
https://github.com/Sharingang/Sharingang-Android/blob/26aa588d0865a3d614c5fe04b1a826d02048c243/app/build.gradle#L47

betrisey (approved):

app/src/main/java/com/example/sharingang/MapFragment.kt
v1ruso: I think it was @SuppressLint("InvalidFragmentVersionForActivityResult"). Missing permission is warning that appears in the code when the permission is not checked right before the feature that requires the permission is called. Since I tried to refactor to have the permission check and the usage of the permission in two separate locations (to avoid having "copy-pasted" code), I had to use this @SuppressLint("MissingPermission")

bafavre: Oh yeah I thought it was already fixed, didn't remember it would be in the next release, my bad.

PR #73, opened 2021-03-23
Add a cache for items

### LINKED ISSUE #36: Offline mode
As an user, I want to be able to find last found items when I'm not connected to the internet.

### LINKED ISSUE #37: Add a cache for items with Room
See #36

cronokirby:
https://user-images.githubusercontent.com/22821947/112180548-f21d7100-8bfb-11eb-84a0-9696bb41af51.mov

Fixes #37 
Fixes #36

The way this works is that you have a cache for the items stored on the user's device. This can be refreshed, as is done when opening the app, or when swiping down on the item list. Refreshing performs an actual request to the underlying store of items (Firebase, in our case), and then saves those items into the local database. This allows the user to see those items, and inspect them, etc., even when offline.

The user can't create new items when offline though. Creating an item will use the underlying store, so actually creating it for the world to see, and then refreshes the list of items.


app/src/main/java/com/example/sharingang/items/CachedItemRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    override fun items(): LiveData<List<Item>> {
+        return itemDao.getAllItems()
+    }
+
+    override suspend fun refreshItems() {
+        // This is necessary, since you want to avoid doing this work on the main thread
+        withContext(Dispatchers.IO) {
+            val newItems = store.getAllItems()
+            itemDao.insert(newItems)
```
betrisey, line 22: Ideally we shouldn't call getAllItems all the time.
Maybe on addItem, just get the the added Item from the store and similarly for the other actions.

betrisey (changes requested):
Nice work :)
I'm just a bit concerned with how much traffic the cache refresh will generate

app/src/main/java/com/example/sharingang/items/Item.kt
```diff
@@ -1,35 +1,46 @@
...
+    @ColumnInfo(name = "id")
     @DocumentId
-    val id: String? = null
+    val id: String? = null,
+
+    /**
+     * The ID we use locally
+     */
+    @PrimaryKey(autoGenerate = true)
+    val localId: Int = 0
```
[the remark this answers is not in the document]
cronokirby: Good idea

app/src/main/java/com/example/sharingang/items/CachedItemRepository.kt
cronokirby: On the other hand, this provides a good occasion to refresh the items automatically, and you need network connection to do these things *anyways*.

I think if we're worried about pulling all the items being too expensive, then we'd need to paginate requests *anyways* at that point, since the initial refresh of the items would also be too slow.

cronokirby: One concern would be latency, but this is hidden because you can go on with your business while you've sent off the action, and thus the refresh, to firebase

betrisey: You're right and maybe when we'll have search, we will just retrieve a few items and update that

app/src/main/java/com/example/sharingang/items/CachedItemRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+    override suspend fun addItem(item: Item): String? {
+        return thenRefresh { store.addItem(item) }
+    }
+
+    override suspend fun getItem(id: String): Item? {
+        return itemDao.getItem(id)
+    }
+
+    override suspend fun getAllItems(): List<Item> {
+        return thenRefresh { store.getAllItems() }
```
betrisey, line 41: Here `store.getAllItems` gets called twice, maybe we should save the result of the first call in the cache

cronokirby: Indeed.

app/src/main/java/com/example/sharingang/ItemsListFragment.kt
```diff
@@ -56,6 +56,15 @@ class ItemsListFragment : Fragment() {
 
         setupNavigation()
 
+        binding.swiperefresh.setOnRefreshListener {
+            viewModel.refresh()
+        }
+        viewModel.refreshing.observe(viewLifecycleOwner, {
+            if (!it) {
+                binding.swiperefresh.isRefreshing = false
+            }
```
v1ruso, line 65: I'm not really sure I understand what does "it" represent?

app/src/main/java/com/example/sharingang/NewItemFragment.kt
```diff
@@ -30,8 +30,15 @@ class NewItemFragment : Fragment() {
...
-            { consumeLocation(fusedLocationCreate, cancellationTokenSource) { updateLocation(it) } }, null)
+        doOrGetPermission(
+            this, Manifest.permission.ACCESS_FINE_LOCATION,
+            {
+                consumeLocation(
+                    fusedLocationCreate,
+                    cancellationTokenSource
+                ) { updateLocation(it) }
+            }, null
+        )
```
v1ruso, line 41: I had to reduce the length of this part (into two lines) to reduce the amount of uncovered code, you might want to put it back as it was if you can't reach the 80% coverage

app/src/main/java/com/example/sharingang/items/CachedItemRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    override suspend fun updateItem(item: Item): Boolean {
+        return thenRefresh { store.updateItem(item) }
+    }
+
+    override suspend fun deleteItem(id: String): Boolean {
```
v1ruso, line 48: For the code duplication, what if we instead have one function editItem that does both actions, depending on a boolean value or enum value passed as parameter?

app/src/main/java/com/example/sharingang/items/InMemoryItemRepository.kt
```diff
@@ -27,10 +27,13 @@ class InMemoryItemRepository @Inject constructor() : ItemRepository {
         return itemsMap.values.toList()
     }
 
-    override fun getAllItemsLiveData(): LiveData<List<Item>> {
+    override fun items(): LiveData<List<Item>> {
         return itemsLiveData
     }
 
+    override suspend fun refreshItems() {
```
v1ruso, line 34: What does this function do, is it intended to be empty?

v1ruso (commented):
I don't understand everything, but it works just fine, well done!

app/src/main/java/com/example/sharingang/items/InMemoryItemRepository.kt
cronokirby: Yeah, the idea is that since this is just mocking the behavior, there's no need to actually refresh anything.

betrisey (approved):

v1ruso (approved):

PR #75, opened 2021-03-24
Add Login (With Account View)

lumapools:
A user can now log in using their Google account. If they need to create a new account, they will be guided through Google's "Create Account" activities, otherwise they can select the Google accounts that are linked to the phone and log in using one of them.

Webp net-resizeimage Webp net-resizeimage (1)

If the user clicks away while they are in Google's login dialog, the login process is cancelled, their status will remain logged out.

Webp net-resizeimage (2) Webp net-resizeimage (3)

This is (for the moment) non-persistent, and when switching activities and coming back, the login information is lost, the user will need to log in again.


app/build.gradle
```diff
@@ -76,11 +76,17 @@ dependencies {
     implementation platform('com.google.firebase:firebase-bom:26.6.0')
     implementation 'com.google.firebase:firebase-firestore-ktx'
 
+    // authentication
+    implementation 'com.google.android.gms:play-services-auth:19.0.0'
+
     implementation "com.google.dagger:hilt-android:$hilt_version"
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation 'com.google.android.gms:play-services-auth:19.0.0'
```
bafavre, line 84: Looks like you have this twice in the file

app/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,8 @@
     package="com.example.sharingang">
 
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.INTERNET"/>
+    <uses-permission android:name="android.permission.WAKE_LOCK"/>
```
bafavre, line 7: Why do you need that permission?

app/src/main/java/com/example/sharingang/AccountFragment.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.google.android.gms.auth.api.signin.GoogleSignInClient
+import com.google.android.gms.auth.api.signin.GoogleSignInOptions
+import com.google.android.gms.common.api.ApiException
+import com.google.android.gms.tasks.OnCompleteListener
+
+
+class AccountFragment : Fragment() {
+
+    private val CLIENT_AUTH_KEY: String =
+        "771023799063-kmve17s9cfu6ckd3kijcv8vdvvdqb58s.apps.googleusercontent.com"
```
bafavre, line 24: Should the key be here in the file?

app/src/main/java/com/example/sharingang/AccountFragment.kt
```diff
@@ -0,0 +1,101 @@
...
+        val toLaunch: Intent = signInClient.signInIntent
+        resultLauncher?.launch(toLaunch)
+        updateUI(accountStatus = AccountStatus.LOGGED_IN, binding) }
+
+    private fun sign_out(signInClient: GoogleSignInClient, binding: FragmentAccountBinding) {
+        signInClient.signOut().addOnCompleteListener(requireActivity()) { updateUI(accountStatus = AccountStatus.LOGGED_OUT, binding) } }
+
+    private fun updateUI(accountStatus: AccountStatus, binding: FragmentAccountBinding) {
+        if(accountStatus == AccountStatus.LOGGED_IN) {
+            (binding.loginButton.visibility) = View.GONE
```
bafavre, line 94: Not that important but you can remove the parentheses.

app/src/main/res/values/strings.xml
```diff
@@ -11,11 +11,19 @@
     <string name="price">Price</string>
     <string name="display_price">$ %.2f</string>
     <string name="item_image_description">Item Image</string>
-    <string name="search_hint">Search...</string>
+    <string name="search_hint" tools:ignore="TypographyEllipsis">Search...</string>
```
bafavre, line 14: Maybe it's better to use the ellipsis character instead of ignoring the warning message.

bafavre (changes requested):
Well done, only a few things to change or maybe you can juste explain to me why you did them :)

app/build.gradle
lumapools: Thanks for pointing it out, I will fix it!

app/src/main/AndroidManifest.xml
lumapools: This was when I was using the initial version, I needed this for the login to be possible, otherwise I was getting weird errors in the console. Fortunately, with the new implementation it works well without wake lock, I forgot to remove it, but will do.

app/src/main/java/com/example/sharingang/AccountFragment.kt
lumapools: The implementations I have seen using it do the same as me. The SHA1 key for authenticating the app is put in oAuth for the project, therefore it is fine to use this OAuth-generated key here as an attribute. Is there a better place you would put it?

app/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,4 @@
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
```
bafavre, line 1: You can remove the tools if you don't use it anymore :)

app/src/main/java/com/example/sharingang/AccountFragment.kt
v1ruso: Isn't such a key supposed to be private? (aka not available directly on the repo?)

bafavre (approved):

betrisey (approved):
Great work!

PR #76, opened 2021-03-25
Add a sold status to items

### LINKED ISSUE #61: Sold status
As a seller, I want to be able to mark an offer as sold so that the buyers don't ask me about it when it has been sold.

### LINKED ISSUE #66: Add a sold field to items
There should be a way to mark an item as sold as well

See #61

### LINKED ISSUE #69: Display sold items differently
See #61

Items that have been sold should look different from normal items.

cronokirby:
Now you can "sell" an item by clicking on a button, and also undo that option. This changes what the item looks like on the main page.

Fixes #61 
Fixes #66 
Fixes #69 :sunglasses:

### DIFF
--- a/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
@@ -40,6 +40,6 @@ class SellItemFragmentTest {
 
         Espresso.onView(ViewMatchers.withText("Sell")).perform(ViewActions.click())
         Espresso.onView(ViewMatchers.withId(R.id.item_list_view_sell_button))
-            .check(ViewAssertions.matches(ViewMatchers.withText("Unsell")))
+            .check(ViewAssertions.matches(ViewMatchers.withText("Resell")))
     }
 }
--- a/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
@@ -1,0 +1,45 @@
+package com.example.sharingang
+
+import androidx.test.espresso.Espresso
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.assertion.ViewAssertions
+import androidx.test.espresso.matcher.ViewMatchers
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+
+@RunWith(AndroidJUnit4::class)
+@HiltAndroidTest
+class SellItemFragmentTest {
+    @get:Rule(order = 0)
+    val hiltRule = HiltAndroidRule(this)
+
+    // We start with the main activity, and then navigate where we want
+    @get:Rule(order = 1)
+    val activityRule = ActivityScenarioRule(MainActivity::class.java)
+
+    private val itemTitle = "T"
+
+    @Test
+    fun anItemCanBeEditedAndSeenOnItemsListFragment() {
+        Espresso.onView(ViewMatchers.withId(R.id.newItemButton)).perform(ViewActions.click())
+        Espresso.onView(ViewMatchers.withId(R.id.newItemPrompt))
+            .check(ViewAssertions.matches(ViewMatchers.withText("New Item")))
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/example/sharingang/items/ItemDatabase.kt
```diff
@@ -5,7 +5,7 @@ import androidx.room.RoomDatabase
 import androidx.room.TypeConverters
 import com.example.sharingang.DateConverter
 
-@Database(entities = [Item::class], version = 1)
+@Database(entities = [Item::class], version = 2)
```
lumapools, line 8: What is the difference between using version 1 and version 2?

app/src/main/res/values/strings.xml
```diff
@@ -14,6 +14,8 @@
     <string name="search_hint">Search...</string>
     <string name="search_go_button">Go</string>
     <string name="latitude">Latitude</string>
+    <string name="sell">Sell</string>
+    <string name="unsell">Unsell</string>
```
lumapools, line 18: I think the words "Sell" and "Unsell" is a bit weird in this context, as upon clicking the button in status "Sell" will not really sell the item right away. For example, I think it would be more user-friendly to use a "Put on sale" or "List for sale" text which describes what the button is doing when the user clicks it, same for the "Unsale" text, something like "Remove from sale" would be a bit more understandable in my opinion.

app/src/main/java/com/example/sharingang/NewItemFragment.kt
```diff
@@ -79,6 +79,7 @@ class NewItemFragment : Fragment() {
                 categoryString = resources.getStringArray(R.array.categories)[binding.categorySpinner.selectedItemPosition],
                 latitude = binding.latitude?.toDoubleOrNull() ?: 0.0,
                 longitude = binding.longitude?.toDoubleOrNull() ?: 0.0,
+                sold = false,
```
lumapools, line 82: Does the attribute `sold` mean "The item has already been bought by another user" or does it mean "The item is not for sale yet"?

lumapools (changes requested):
Nice implementation! Please see my comments for clarification requests and perhaps changing some text values for better user-friendlyness.

app/src/main/java/com/example/sharingang/items/ItemDatabase.kt
cronokirby: So like, when you change the database schema, you're supposed to bump up the version, so it knows how to handle old data.

app/src/main/res/values/strings.xml
cronokirby: Indeed, but maybe we could find something shorter?

app/src/main/java/com/example/sharingang/NewItemFragment.kt
cronokirby: `sold = true` means that the item has been bought by another user, and is no longer available.

app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
```diff
@@ -0,0 +1,45 @@
...
+            ViewActions.typeText(itemTitle),
+            ViewActions.closeSoftKeyboard()
+        )
+        val buttonCreate = Espresso.onView(ViewMatchers.withId(R.id.createItemButton))
+        buttonCreate.check(ViewAssertions.matches(ViewMatchers.withText("Create Item")))
+        buttonCreate.perform(ViewActions.click())
+
+        Espresso.onView(ViewMatchers.withText("Sell")).perform(ViewActions.click())
+        Espresso.onView(ViewMatchers.withId(R.id.item_list_view_sell_button))
+            .check(ViewAssertions.matches(ViewMatchers.withText("Unsell")))
```
v1ruso, line 43: Same as last week, it'd be cleaner to import the functions (just onView instead of Espresso.onView).

v1ruso (approved):

app/src/main/res/values/strings.xml
lumapools: I looked up synonyms but nothing was explicit/understandable enough :(, do you have any ideas?

cronokirby: Watch this:
```suggestion
    <string name="unsell">Resell</string>
```

cronokirby: Look ma, I didn't even open Android Studio

lumapools (approved):

app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
cronokirby, line 43:
```suggestion
            .check(ViewAssertions.matches(ViewMatchers.withText("Resell")))
```

PR #78, opened 2021-03-25
55 search

### LINKED ISSUE #33: Create a view to search for items

### LINKED ISSUE #55: Implement name search functionality

### LINKED ISSUE #77: Search by category
As a Buyer, I want to filter the items by categories in order to easily find what I want.

Superdoude1999:
Added search functionality, implementation goes as follows:
A search with a string will (for now) only search for items who's title contains the given string. Plan to improve by splitting the string into multiple words if necessary and searching for each word individually. For the search there are a few options:
- If user only enters a name, search will ignore category search.
- If user only enters a category, search will ignore name search.
- If user enters both a name and a category, search will only return items that match both criteria.
There is also a clear search button which will reset the search and remove and current results from the view.
Closes #55 
Closes #77 
Closes #33


app/src/main/java/com/example/sharingang/ItemsListFragment.kt
```diff
@@ -68,6 +64,10 @@ class ItemsListFragment : Fragment() {
         return binding.root
     }
 
+    private fun gotoNewItem(view: View){
```
v1ruso, line 67: It'd be nicer with "goToNewItem"

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
```diff
@@ -1,114 +1,162 @@
...
+                for(item in allItemsSet){
+                    if(item.category != categoryID){
+                        categoryResults.remove(item)
+                    }
+                }
+            }
+
+            if(searchName.isNotEmpty()){
+                for(item in allItemsSet){
+                    if(!item.title.toLowerCase().contains(searchName!!.toLowerCase())){
```
v1ruso, line 95: See Cirrus Warning: shouldn't use the default locale for toLowerCase

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
```diff
@@ -1,114 +1,162 @@
...
+                        nameResults.remove(item)
```
v1ruso, line 96: I'm not really convinced by the algorithm. You're taking all the items and removing the ones that don't contain a certain string. Wouldn't it be better to start with an empty list and add items that have a certain string in their title?

app/src/main/res/layout/fragment_search.xml
```diff
@@ -9,22 +9,43 @@
         android:orientation="vertical">
 
         <EditText
-            android:id="@+id/sflSearchText"
-            android:layout_width="231dp"
-            android:layout_height="50dp"
+            android:id="@+id/searchText"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
             android:ems="10"
```
v1ruso, line 15: Please specify the inputType

v1ruso (changes requested):

betrisey (changes requested):

betrisey (approved):

cronokirby:
BTW, I'd recommend adding a `.gitattributes` file at the root of the repo, containing:

```
* text=auto
```

And then running:

```
git add --renormalize .
```

To fix all of the line endings.

Then you can commit and push that.

Just an FYI

v1ruso (approved):

PR #88, opened 2021-03-31
59 display items

### LINKED ISSUE #59: Add a way to display the nearby items on the map

v1ruso:
Fix #59 
I added a way to display the items on the map (in red, the current location of the user, in blue, the items location). 
before_click
For the moment, it displays all the items currently stored, but we could filter the items by distance, or by category, for example.

After clicking on an item marker, a description of the item appears on screen. 
after_click

We could display a full description of the items instead, like when clicking on an item on the main page.


bafavre (approved):
All good, well done!

PR #90, opened 2021-04-01
Replace buttons with navigation

### LINKED ISSUE #83: Navigation
As a User, I want to easily be able to navigate to different sections of the app, to make it easier to use

### LINKED ISSUE #87: Implement navigation
See #83

cronokirby:
Fixes #87
Fixes #83

This removes the ugly buttons, replacing things with a standard navigation drawer:
image
image
image

I also had to edit literally every single test to use this navigation now :P

### DIFF
--- a/app/src/androidTest/java/com/example/sharingang/AccountFragmentTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/AccountFragmentTest.kt
@@ -1,24 +1,13 @@
 package com.example.sharingang
 
-import android.Manifest
-import android.content.Intent
 import androidx.test.espresso.Espresso.onView
-import androidx.test.espresso.Espresso.pressBack
-import androidx.test.espresso.action.ViewActions.*
 import androidx.test.espresso.assertion.ViewAssertions.matches
-import androidx.test.espresso.intent.Intents
-import androidx.test.espresso.intent.Intents.intended
-import androidx.test.espresso.intent.matcher.IntentMatchers.toPackage
 import androidx.test.espresso.matcher.ViewMatchers.*
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
-import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
-import androidx.test.rule.GrantPermissionRule
-import androidx.test.uiautomator.UiDevice
 import dagger.hilt.android.testing.HiltAndroidRule
 import dagger.hilt.android.testing.HiltAndroidTest
 import org.hamcrest.Matchers.not
-import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -35,25 +24,9 @@ class AccountFragmentTest {
 
     @Test
     fun logoutButtonDoesNotAppearWhenLoggedOut() {
-        onView(withId(R.id.gotoAccount)).check(matches(withText("Account")))
-        onView(withId(R.id.gotoAccount)).perform(click())
+        navigate_to(R.id.accountFragment)
         onView(withId(R.id.loginButton)).check(matches(withText("Log In")))
         onView(withId(R.id.logoutButton)).check(matches(not(isDisplayed())))
         onView(withId(R.id.account_status)).check(matches(withText("Status: Logged Out")))
     }
-
-    /*
-    @Test
-    fun loginClick() {
-        Thread.sleep(5000)
-        val device: UiDevice = UiDevice.getInstance(getInstrumentation())
-        onView(withId(R.id.gotoAccount)).perform(click())
-        onView(withId(R.id.loginButton)).perform(click())
-        device.pressBack()
-        onView(withId(R.id.loginButton)).check(matches(withText("Log In")))
-    }
-    */
-
-
-
 }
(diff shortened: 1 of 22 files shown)


app/src/androidTest/java/com/example/sharingang/SearchFragmentTest.kt
```diff
@@ -22,28 +20,27 @@ import java.lang.Exception
...
     val hiltRule = HiltAndroidRule(this)
 
-    val firstTitle : String = "Harry Potter"
-    val secondTitle : String = "Inception"
-    val thirdTitle : String = "iPad"
-    val fourthTitle : String = "Harry Potter 2"
+    private val firstTitle: String = "Harry Potter"
+    private val secondTitle: String = "Inception"
+    private val thirdTitle: String = "iPad"
+    private val fourthTitle: String = "Harry Potter 2"
```
v1ruso, line 29: Well done

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -1,7 +1,39 @@
...
+            supportFragmentManager.findFragmentById(R.id.myNavHostFragment) as NavHostFragment
+        val navController = navHostFragment.navController
+
+        val drawerLayout = findViewById<DrawerLayout>(R.id.drawer_layout)
+        val appBarConfiguration = AppBarConfiguration(navController.graph, drawerLayout)
+        findViewById<Toolbar>(R.id.toolbar)
+            .setupWithNavController(navController, appBarConfiguration)
+        findViewById<NavigationView>(R.id.nav_view)
+            .setupWithNavController(navController)
+    }
```
v1ruso, line 31: I see you are using multiple findViewById, could we use a binding instead?

app/src/main/res/values-night/themes.xml
```diff
@@ -1,6 +1,6 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
     <!-- Base application theme. -->
-    <style name="Theme.Sharingang" parent="Theme.MaterialComponents.DayNight.DarkActionBar">
+    <style name="Theme.Sharingang" parent="Theme.MaterialComponents.DayNight.NoActionBar">
```
v1ruso, line 3: What does this do?

app/src/androidTest/java/com/example/sharingang/AccountFragmentTest.kt
```diff
@@ -35,25 +24,9 @@ class AccountFragmentTest {
 
     @Test
     fun logoutButtonDoesNotAppearWhenLoggedOut() {
-        onView(withId(R.id.gotoAccount)).check(matches(withText("Account")))
-        onView(withId(R.id.gotoAccount)).perform(click())
+        navigate_to(R.id.accountFragment)
```
v1ruso, line 27: Must have been a pain to update all the tests x) thank you

v1ruso (changes requested):
It works just fine, well done

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -1,7 +1,39 @@
...
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        val navHostFragment =
+            supportFragmentManager.findFragmentById(R.id.myNavHostFragment) as NavHostFragment
+        val navController = navHostFragment.navController
+        return item.onNavDestinationSelected(navController) || super.onOptionsItemSelected(item)
```
v1ruso, line 37: What does this do? I'm a bit confused with the result value, what does the boolean represent?

app/src/main/res/values-night/themes.xml
cronokirby: It removes the action bar that used to say Sharingang at the top of the app

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: You should return true when you've handled the event, so that Android knows to close the drawer, and other things like that.

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: For some reason like, the tests fail when running without findViewById?

I think Android needs something specific to setup navigation. I used a binding instead of findViewById, and it ended up even making the menu thingy disappear for some reason

v1ruso: Ok, then all good then

v1ruso (approved):

app/src/main/java/com/example/sharingang/ItemsListFragment.kt
```diff
@@ -74,29 +57,5 @@ class ItemsListFragment : Fragment() {
...
-        view.findNavController().navigate(
-                ItemsListFragmentDirections.actionItemsListFragmentToSearchFragment()
-        )
-    }
-
-    fun goToAccount(view: View) {
-        view.findNavController().navigate(
-                ItemsListFragmentDirections.actionItemsListFragmentToAccountFragment()
-        )
-    }
```
betrisey, line 100: Nice, we no longer need all of methods :)

betrisey (approved):
Very good!

cronokirby:
Thankya!

PR #91, opened 2021-04-01
Taking pictures

### LINKED ISSUE #60: Taking Pictures
As a Seller, I want to be able to take a picture using the app, to make including an image description easy

### LINKED ISSUE #65: Add an Image from the Camera to an Item

bafavre:
Fixes #60
Fixes #65 

We can now use the camera to take pictures of items and display them.


app/src/main/java/com/example/sharingang/utils/ImageAccess.kt
```diff
@@ -0,0 +1,188 @@
...
+        ) { isGranted: Boolean ->
+            if (!isGranted)
+                Toast.makeText(
+                    registry,
+                    "$permission permission successfully denied. Feature is disabled.",
+                    Toast.LENGTH_LONG
+                ).show()
+            when (permission) {
+                Manifest.permission.CAMERA -> cameraPermissionGranted = isGranted
+                Manifest.permission.READ_EXTERNAL_STORAGE -> storagePermissionGranted = isGranted
```
v1ruso, line 97: Didn't we say that we should avoid the use of boolean values?

app/src/main/java/com/example/sharingang/utils/ImageAccess.kt
```diff
@@ -0,0 +1,188 @@
...
+                Toast.makeText(context, rationale, Toast.LENGTH_LONG).show()
+            }
+            else -> {
+                when (permission) {
+                    Manifest.permission.READ_EXTERNAL_STORAGE -> requestStoragePermissionLauncher.launch(
+                        permission
+                    )
+                    Manifest.permission.CAMERA -> requestCameraPermissionLauncher.launch(permission)
+                }
+            }
```
v1ruso, line 127: This whole block looks similar to what is in the file PermissionRequest. Can't we refactor?

app/src/main/java/com/example/sharingang/utils/ImageAccess.kt
bafavre: The problem we had was not that we were using a boolean, it was that we were not checking the permissions everytime, but here it is done. Maybe I could do without it, but not sure.

app/src/main/java/com/example/sharingang/utils/ImageAccess.kt
bafavre: I can't really, first because I don't have a fragment in ImageAccess (but I can surely make it take one) but most importantly because my callback isn't a ()  -> Unit, but can be a (String) -> Uri or (Uri) -> Boolean

betrisey (approved):
Nice, the code looks good :)
The only thing I see that can be improved is when we click on "Take Picture" for the first time and grant the permission, we have to click a second time on the button. If it's not too complicated, it could be nice if the camera is opened when the permission is granted.

app/src/main/java/com/example/sharingang/utils/ImageAccess.kt
v1ruso: Fair enough

v1ruso (approved):
Well done!

PR #93, opened 2021-04-04
Add login persistence (Locally + Firebase)

lumapools:
The app will, from now on, remember which user is logged in (even if the user quits the app) and link the account with Firebase (add the public information of the user to the database) and the User Profile Fragment. The picture of the user is shown on the profile.
Left: Account page (login action + status), 
Right: User Profile (with profile picture)
Webp net-resizeimage (4) Webp net-resizeimage (5)


app/src/main/java/com/example/sharingang/ItemsListFragment.kt
```diff
@@ -42,8 +43,20 @@ class ItemsListFragment : Fragment() {
...
+                getString(R.string.preference_user_info),
+                Context.MODE_PRIVATE
+            )
             val action =
+                if (sharedPreferences == null) {
                     ItemsListFragmentDirections.actionItemsListFragmentToUserProfileFragment("test")
+                } else {
+                    ItemsListFragmentDirections.actionItemsListFragmentToUserProfileFragment(
+                        sharedPreferences.getString(
+                            getString(R.string.account_firebase_uid), "test"
```
bafavre, line 56: Should we keep "test" as a default string? I don't know what else to put but maybe test isn't the best.

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -33,10 +33,11 @@ class UserProfileFragment : Fragment() {
         viewModel.setUser(args.userId)
         binding.viewModel = viewModel
 
+        /*
         lifecycleScope.launch(Dispatchers.IO) {
             viewModel.createTestUser(args.userId)
         }
-
+        */
```
bafavre, line 40: Comments should be removed

app/src/main/java/com/example/sharingang/UserProfileViewModel.kt
```diff
@@ -27,7 +27,7 @@ class UserProfileViewModel @Inject constructor(
     fun setUser(userId: String) {
         _userId.postValue(userId)
     }
-
+    /*
     // Only for testing
     suspend fun createTestUser(userId: String) {
         if (userRepository.get(userId) == null) {
```
bafavre, line 33: Maybe remove the comments

bafavre (changes requested):
Good job, only a few remarks :)
Also maybe consider using DataStore instead of SharedPreferences : https://developer.android.com/topic/libraries/architecture/datastore

app/src/main/java/com/example/sharingang/AccountFragment.kt
```diff
@@ -97,4 +145,78 @@ class AccountFragment : Fragment() {
...
+                    lifecycleScope.launch(Dispatchers.IO) {
+                        userRepository.add(
+                            User(
+                                id = user?.uid,
+                                name = user?.displayName!!,
+                                profilePicture = user.photoUrl!!.toString()
+                            )
+                        )
+                    }
+                    editor.putString(getString(R.string.account_firebase_uid), user?.uid)
```
betrisey, line 185:
```suggestion
                if (task.isSuccessful && auth.currentUser != null) {
                    val user = auth.currentUser!!
                    lifecycleScope.launch(Dispatchers.IO) {
                        userRepository.add(
                            User(
                                id = user.uid,
                                name = user.displayName!!,
                                profilePicture = user.photoUrl?.toString()
                            )
                        )
                    }
                    editor.putString(getString(R.string.account_firebase_uid), user.uid)
```
I think it's better to check if the user is not null before. The code becomes a bit simpler inside and we avoid adding null user to the database.

app/src/main/java/com/example/sharingang/ItemsListFragment.kt
betrisey, line 56: If the user is not logged in, I think we shouldn't go to the user profile at all, maybe display a message with a Toast or Snackbar.
Also for the tests, I'll try to see if we can use dependency injection to get a fake user ID for the tests.

betrisey (changes requested):

betrisey (approved):

bafavre (approved):

PR #95, opened 2021-04-10
Update README

lumapools:
We don't only need the Firestore emulator now, we also need the Auth emulator for authentication with Firebase.


betrisey (approved):

Superdoude1999 (approved):
Much wow

PR #96, opened 2021-04-10
Add deeplink support

betrisey:
- Add a share button on the detailed view, it generates and share a link
- When the link is clicked, the app is opened and the item displayed


app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -42,10 +50,35 @@ class MainActivity : AppCompatActivity(R.layout.activity_main) {
...
 
+    /*
+    We don't have an option menu yet
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         val navController = getNavController()
         return item.onNavDestinationSelected(navController) || super.onOptionsItemSelected(item)
     }
-}
\ No newline at end of file
+    */
```
bafavre: Should we keep commented code?

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
```diff
@@ -139,7 +133,7 @@ class ItemsViewModel @Inject constructor(
             OBSERVABLES.SEARCH_RESULTS -> searchResults
         }
         observable.observe(LifeCycleOwner, {
-            it?.let { adapter.submitList(it) }
+            adapter.submitList(it)
```
bafavre, line 136: What happens if "it" is null here? Everything is good or "it" cannot be null or ?

bafavre (approved):
Well done, I just had a few questions in the code :)

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
betrisey: "it" is a List<Item> and not a List<Item>?
So Android Studio suggested to remove the check

app/src/main/java/com/example/sharingang/MainActivity.kt
betrisey: I think I'll remove it, thanks

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
bafavre: Oh yeah then makes sense, all good!

app/src/main/java/com/example/sharingang/MapFragment.kt
```diff
@@ -138,11 +138,9 @@ class MapFragment : Fragment(), OnMapReadyCallback {
             map?.animateCamera(CameraUpdateFactory.newLatLng(marker.position))
             if (marker != lastLocationMarker) {
                 val temp: Item = marker.tag as Item
-                viewModel.onViewItem(temp)
                 this.findNavController().navigate(
-                    MapFragmentDirections.actionMapFragmentToDetailedItemFragment()
+                    MapFragmentDirections.actionMapFragmentToDetailedItemFragment(temp)
```
Superdoude1999, line 142: not important, but maybe find a more suitable name than temp

Superdoude1999 (approved):

PR #99, opened 2021-04-12
Change to firebase auth

lumapools:
Most of these changes have been made by @betrisey.

Authentication method is now changed to FirebaseAuth. There are no changes in the UI whatsoever.
This changes makes the code more concise and easier to understand.


betrisey (approved):

bafavre (approved):
Maybe check the warnings of Cirrus (for example the unused strings) but otherwise all good! :)

PR #104, opened 2021-04-15
Linking items to users

### LINKED ISSUE #82: Linking items with User
As a Buyer, I want to see who the Seller of an item is, to learn more about them and the item

### LINKED ISSUE #101: Display the User that added an item

bafavre:
Fixes #101 
Fixes #82 

Now if we are connected when adding an item, the item is linked to us. We can also see the profile of the person that posted the item by clicking on it in the detailed view.

### DIFF
--- a/app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
@@ -9,6 +9,7 @@ import android.view.ViewGroup
 import androidx.databinding.DataBindingUtil
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
+import androidx.navigation.findNavController
 import androidx.navigation.fragment.navArgs
 import com.example.sharingang.databinding.FragmentDetailedItemBinding
 import com.example.sharingang.items.Item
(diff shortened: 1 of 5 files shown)


Superdoude1999 (approved):
ok

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -27,9 +32,20 @@ class DetailedItemFragment : Fragment() {
             DataBindingUtil.inflate(inflater, R.layout.fragment_detailed_item, container, false)
 
         binding.item = args.item
-
         binding.shareButton.setOnClickListener { shareItem() }
 
+        viewModel.setUser(args.item.userId)
+        viewModel.user.observe(viewLifecycleOwner, { user ->
+            binding.username = "Posted by " + user?.name
```
cronokirby, line 39:
```suggestion
            binding.username = "Posted by ${user?.name}"
```

I prefer using string interpolation, but that's kind of a personal preference I guess

app/src/main/res/layout/fragment_detailed_item.xml
```diff
@@ -48,6 +51,14 @@
             android:textSize="18sp"
             tools:text="Here goes the description of the item." />
 
+        <TextView
+            android:id="@+id/itemPostedBy"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:textSize="18sp"
+            android:text="@{username}"
+            tools:text="Posted by TestUser" />
```
cronokirby, line 60: Maybe we should make this a string resource?

cronokirby (changes requested):
Just a handful of minor things, otherwise, great job!

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: Oh I thought I actually did that because Android Studio complained, but I guess I somehow reverted it.

app/src/main/res/layout/fragment_detailed_item.xml
bafavre: It's a tools text, so I don't know if we should do that, I think we haven't done it for any of the others.

cronokirby: Yeah, I guess it's fine

cronokirby (approved):
Cool Beans!

PR #107, opened 2021-04-15
Fix item image after edition

### LINKED ISSUE #106: Fix item image not displaying after edition

bafavre:
Fixes #106

Before the item image URI would be lost if we would edit the item without changing the image.

### DIFF
--- a/app/src/main/java/com/example/sharingang/EditItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/EditItemFragment.kt
@@ -77,7 +77,7 @@ class EditItemFragment : Fragment() {
                     categoryString = resources.getStringArray(R.array.categories)[binding.categorySpinner.selectedItemPosition],
                     latitude = binding.latitude?.toDoubleOrNull() ?: 0.0,
                     longitude = binding.longitude?.toDoubleOrNull() ?: 0.0,
-                    imageUri = imageUri?.toString()
+                    imageUri = if(imageUri != null) imageUri.toString() else existingItem.imageUri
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/example/sharingang/EditItemFragment.kt
```diff
@@ -77,7 +77,7 @@ class EditItemFragment : Fragment() {
                     categoryString = resources.getStringArray(R.array.categories)[binding.categorySpinner.selectedItemPosition],
                     latitude = binding.latitude?.toDoubleOrNull() ?: 0.0,
                     longitude = binding.longitude?.toDoubleOrNull() ?: 0.0,
-                    imageUri = imageUri?.toString()
+                    imageUri = if(imageUri != null) imageUri.toString() else existingItem.imageUri
```
cronokirby, line 80:
```suggestion
                    imageUri = imageUri?.toString() ?: existingItem.imageUri
```

Pretty sure that this works, and is more idiomatic kotlin

cronokirby (changes requested):

cronokirby:
Nice catch otherwise!

betrisey (approved):

app/src/main/java/com/example/sharingang/EditItemFragment.kt
bafavre: Yeah you're right, I tried to do something like that but failed at first, thanks!

cronokirby (approved):
Cool beans!

PR #108, opened 2021-04-15
Allow seeing items in the detailed fragment view

### LINKED ISSUE #103: Add the ability to see item images
See #102

cronokirby:
Fixes #103

image

### DIFF
--- a/app/src/main/res/layout/fragment_detailed_item.xml
+++ b/app/src/main/res/layout/fragment_detailed_item.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<layout xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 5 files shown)


betrisey (approved):

app/src/main/res/layout/fragment_detailed_item.xml
```diff
@@ -24,6 +24,13 @@
             android:textSize="30sp"
             tools:text="Title" />
 
+        <ImageView
+            android:id="@+id/edit_item_image"
```
bafavre, line 28: Maybe change the id so it's not the same as in the edit fragment, or it would make sense to change the id so it doesn't contain edit.

bafavre (changes requested):
Good job!
Just a small detail to take into consideration.

app/src/main/res/layout/fragment_detailed_item.xml
cronokirby: Yeah, that's a good point!

bafavre (approved):

PR #110, opened 2021-04-16
Profile picture

lumapools:
Add possibility for user to update their profile picture. As the user would like to be recognized by other users (buyers by sellers or sellers by buyers), they would probably prefer to use a picture of their face in order for the meetup to go better as if they were seeing each other's face for the first time.

Webp net-resizeimage

Using the Camera:

Webp net-resizeimage (1) Webp net-resizeimage (2)

Using the Gallery:

Webp net-resizeimage (3) Webp net-resizeimage (4)

The apply button will udpate the changes, so that after the user chooses/takes the picture and thinks it looks good on the app, they can submit their changes and their profile picture gets updated both locally and in the database.


app/src/androidTest/java/com/example/sharingang/UserProfileFragmentTest.kt
```diff
@@ -21,11 +26,51 @@ class UserProfileFragmentTest {
...
     val activityRule = ActivityScenarioRule(MainActivity::class.java)
 
+    @get:Rule(order = 2)
+    val permissionRule: GrantPermissionRule = GrantPermissionRule.grant(
+        Manifest.permission.READ_EXTERNAL_STORAGE,
+        Manifest.permission.CAMERA
+    )
+
     @Test
     fun canOpenSearchFragment() {
```
v1ruso, line 36: I think the name is wrong here?

app/src/androidTest/java/com/example/sharingang/UserProfileFragmentTest.kt
```diff
@@ -21,11 +26,51 @@ class UserProfileFragmentTest {
...
+    fun applyButtonApplies() {
+        val device: UiDevice = UiDevice.getInstance(getInstrumentation())
+        navigate_to(R.id.userProfileFragment)
+        onView(withId(R.id.btn_open_gallery)).perform(click())
+        device.pressBack()
+        onView(withId(R.id.btnApply)).check(matches(isDisplayed()))
+        UserProfileFragment.isTestCase = true
+        onView(withId(R.id.btnApply)).perform(click())
+        onView(withId(R.id.btnApply)).check(matches(not(isDisplayed())))
+        UserProfileFragment.isTestCase = false
```
v1ruso, line 74: Isn't there a way to test the class without using attributes made specifically for testing purposes?

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -1,50 +1,123 @@
...
+            }
+        }
+    }
+
+    private fun getAction(button: Button) {
+        when (button) {
+            binding.btnOpenCamera -> imageAccess.openCamera()
+            binding.btnOpenGallery -> imageAccess.openGallery()
+            binding.btnApply -> {
+                imageUri = if (!isTestCase) imageAccess.getImageUri() else Uri.parse(TESTURI)
```
v1ruso, line 106: As mentioned in the test, I am not a huge fan of this testCase variable. Isn't there a way to avoid that?

v1ruso (changes requested):

app/src/androidTest/java/com/example/sharingang/UserProfileFragmentTest.kt
lumapools: Indeed

app/src/androidTest/java/com/example/sharingang/UserProfileFragmentTest.kt
lumapools: I can remove it but for now I have not found any other way to test the gallery or the camera's response, even with Intent stubbing there are too many unknowns because we are using a more or less custom interface for image support. I guess the options would be either to keep it as it is, or remove it but the coverage would go down.

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -1,51 +1,179 @@
...
 
+
+
 @AndroidEntryPoint
 class UserProfileFragment : Fragment() {
     private val userViewModel: UserProfileViewModel by viewModels()
     private val args: UserProfileFragmentArgs by navArgs()
     private lateinit var binding: UserProfileFragmentBinding
+    // This is the currently logged in user
+    private var currentUserId: String? = ""
```
v1ruso, line 33: Do we need "String?" if it is initialized to a non-null value?

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -1,51 +1,179 @@
...
+            topInfoText.visibility = View.GONE
+        }
+    }
+
+    private fun isAuthUserDisplayedUser(): Boolean {
+        val shownUserId = args.userId
+        return shownUserId == null || shownUserId == currentUserId
+    }
+
+    private fun setupPfpAndNameVisibility() {
```
v1ruso, line 128: What is "Pfp"? Might rename it to make it clearer

v1ruso (approved):
Well done

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
Superdoude1999: Maybe just init to null so when we check that a user is logged in it actually triggers the null checks

app/src/androidTest/java/com/example/sharingang/FakeCurrentUserProvider.kt
```diff
@@ -21,6 +21,10 @@ class FakeCurrentUserProvider @Inject constructor(
         return fakeUser.id
     }
 
+    override fun getCurrentUserEmail(): String? {
+        return "test user email"
```
betrisey, line 25: Maybe use something with a correct e-mail format: "test-user@example.com"

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -1,51 +1,179 @@
...
+            binding.gallerycameraholder,
+            binding.nameText,
+            binding.textEmail,
+            binding.applyholder
+        )
+        for(view: View in fields) {
+            view.visibility = View.GONE
+        }
+        binding.textEmail.text = getString(R.string.text_email_unavailable)
+    }
```
Superdoude1999, line 87: Maybe don't need to create a list for that (but maybe code climate is not nice)

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
Superdoude1999: I concur

Superdoude1999 (commented):
Good job sir.

Superdoude1999 (approved):

betrisey (approved):
Great :)
just minor details to change

Issue #105, opened 2021-04-15
Wish List


Superdoude1999:
As a user, I want to have a list of all items I want to buy before buying them, so that I can see if their prices change or if they get sold.

PR #114, opened 2021-04-16
105 wishlist

Superdoude1999:
Wishlist attribute added to User, so each user has a wishlist.
Added the "Add to wishlist" button to the detailed item view. Moved the navigation setup for items to the viewModel. 
Filled in the UserProfileViewModel to implement the wishlist such that it updates immediately when adding or removing items as well as when changing users.
Still need to finish adding a way to view it by either creating a new fragment or adding it to user profile fragment but can't make it work yet (tried with userProfile and a recycler view). Also still need to write tests for it.
Issues:
#105 
#111 
#112 
#113

### DIFF
--- a/app/src/main/java/com/example/sharingang/DateConverter.kt
+++ b/app/src/main/java/com/example/sharingang/DateConverter.kt
@@ -1,7 +1,16 @@
 package com.example.sharingang
 
+import android.os.Build
+import androidx.annotation.RequiresApi
+import androidx.lifecycle.MutableLiveData
 import androidx.room.TypeConverter
+import com.example.sharingang.items.Item
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import kotlinx.coroutines.GlobalScope
+import java.lang.StringBuilder
 import java.util.*
+import kotlin.collections.ArrayList
 
 /**
  * A utility object needed to store Date fields in our Room entities.
@@ -21,5 +30,22 @@ object DateConverter {
             return null
         }
         return date.time
+    }
+
+
+    @RequiresApi(Build.VERSION_CODES.N)
+    @TypeConverter
+    fun fromList(wishlist: MutableLiveData<List<String>>): String{
+        val returnString : StringJoiner = StringJoiner(",")
+        wishlist!!.value!!.map { str ->
+            returnString.add(str)
+        }
+        return returnString.toString()
+    }
+
+    @TypeConverter
+    fun toList(data: String): MutableLiveData<MutableList<String>>? {
+        val list = data.split(',')
+        return MutableLiveData(ArrayList(list))
     }
 }
--- a/app/src/main/java/com/example/sharingang/DateConverter.kt
+++ b/app/src/main/java/com/example/sharingang/DateConverter.kt
@@ -32,20 +32,16 @@ object DateConverter {
         return date.time
     }
 
-
-    @RequiresApi(Build.VERSION_CODES.N)
     @TypeConverter
-    fun fromList(wishlist: MutableLiveData<List<String>>): String{
-        val returnString : StringJoiner = StringJoiner(",")
-        wishlist!!.value!!.map { str ->
-            returnString.add(str)
+    fun toList(stringList: String?): List<String?>{
+        if(stringList == null){
+            return ArrayList()
         }
-        return returnString.toString()
+        return stringList.split(',')
     }
 
     @TypeConverter
-    fun toList(data: String): MutableLiveData<MutableList<String>>? {
-        val list = data.split(',')
-        return MutableLiveData(ArrayList(list))
+    fun fromList(list : List<String?>): String {
+        return list.joinToString(separator = ",")
     }
 }
--- a/app/src/main/java/com/example/sharingang/DateConverter.kt
+++ b/app/src/main/java/com/example/sharingang/DateConverter.kt
@@ -2,7 +2,6 @@ package com.example.sharingang
 
 import androidx.room.TypeConverter
 import java.util.*
-import kotlin.collections.ArrayList
 
 /**
  * A utility object needed to store Date fields in our Room entities.
--- a/app/src/main/java/com/example/sharingang/DateConverter.kt
+++ b/app/src/main/java/com/example/sharingang/DateConverter.kt
@@ -1,14 +1,6 @@
 package com.example.sharingang
 
-import android.os.Build
-import androidx.annotation.RequiresApi
-import androidx.lifecycle.MutableLiveData
 import androidx.room.TypeConverter
-import com.example.sharingang.items.Item
-import com.google.gson.Gson
-import com.google.gson.reflect.TypeToken
-import kotlinx.coroutines.GlobalScope
-import java.lang.StringBuilder
 import java.util.*
 import kotlin.collections.ArrayList
 
--- a/app/src/main/java/com/example/sharingang/DateConverter.kt
+++ b/app/src/main/java/com/example/sharingang/DateConverter.kt
@@ -24,16 +24,5 @@ object DateConverter {
         return date.time
     }
 
-    @TypeConverter
-    fun toList(stringList: String?): List<String?>{
-        if(stringList == null){
-            return ArrayList()
-        }
-        return stringList.split(',')
-    }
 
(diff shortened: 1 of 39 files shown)


app/src/main/java/com/example/sharingang/UserProfileViewModel.kt
```diff
@@ -24,7 +39,53 @@ class UserProfileViewModel @Inject constructor(
             }
         }
 
+    fun refreshListUI(viewModel: ItemsViewModel){
+        viewModelScope.launch(Dispatchers.IO){
+            if(userId.value!= null){
+                val itemList = ArrayList<Item>()
+                for(str in userRepository.get(userId.value!!)!!.wishlist){
```
v1ruso, line 46: I think you don't need the "!!" for userId.value!!, it's null checked at line 44

v1ruso (commented):
Cool:)

app/src/main/java/com/example/sharingang/UserProfileViewModel.kt
bafavre: I think he told me even though he null checks it, android studio was complaining here without it

app/src/main/java/com/example/sharingang/DateConverter.kt
```diff
@@ -22,4 +23,17 @@ object DateConverter {
...
+        if(stringList == null){
+            return ArrayList()
+        }
+        return stringList.split(',')
+    }
+
+    @TypeConverter
+    fun fromList(list : List<String?>): String {
+        return list.joinToString(separator = ",")
+    }
```
bafavre, line 38: I don't think the list converter should be in the DateConverter file.

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -46,7 +53,10 @@ class DetailedItemFragment : Fragment() {
             binding.detailedItemImage.setImageURI(Uri.parse(it))
         }
 
+        initiateWishlistButton(binding)
+
         binding.shareButton.setOnClickListener { shareItem() }
+        binding.addToWishlist.setOnClickListener { updateWishlist(binding) }
```
bafavre, line 59: Can't you put all wishlist button related code in the initiateWishlistButton method?

app/src/main/java/com/example/sharingang/UserProfileViewModel.kt
```diff
@@ -1,20 +1,35 @@
...
 
 @HiltViewModel
 class UserProfileViewModel @Inject constructor(
+    currentUserProvider: CurrentUserProvider,
+    private val itemRepository: ItemRepository,
     private val userRepository: UserRepository
 ) : ViewModel() {
 
-    private val _userId = MutableLiveData<String?>()
+    private val _userId = MutableLiveData<String?>(currentUserProvider.getCurrentUserId())
```
bafavre, line 22: When we go to the user profile, it's not necessarily only the current user we'll see. We can also see other users and that's why we use the setUser method to change it.

app/src/main/java/com/example/sharingang/UserProfileViewModel.kt
```diff
@@ -1,20 +1,35 @@
...
+    val userId: LiveData<String?>
+        get() = _userId
```
bafavre, line 24: We usually get the whole user instead of just the id with the val user, so no need to have a val userId here in my opinion.

app/src/main/java/com/example/sharingang/users/User.kt
```diff
@@ -4,11 +4,12 @@ import androidx.room.ColumnInfo
 import androidx.room.Entity
 import androidx.room.Index
 import androidx.room.PrimaryKey
-import java.util.Date
-
 import com.google.firebase.firestore.DocumentId
 import com.google.firebase.firestore.Exclude
 import com.google.firebase.firestore.ServerTimestamp
+import java.util.*
```
bafavre, line 10: Why importing the whole java.util package?

app/src/main/res/layout/fragment_wishlist_view.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+
+
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
```
bafavre, line 7: You can remove it if you don't use it.

app/src/main/res/layout/user_profile_fragment.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<layout xmlns:android="http://schemas.android.com/apk/res/android">
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
```
bafavre, line 3: Same here, can be removed.

app/src/main/res/navigation/navigation.xml
```diff
@@ -92,4 +95,15 @@
         android:id="@+id/accountFragment"
         android:name="com.example.sharingang.AccountFragment"
         android:label="Account" />
+    <fragment
+        android:id="@+id/wishlistViewFragment"
+        android:name="com.example.sharingang.WishlistViewFragment"
+        android:label="WishlistViewFragment" >
```
bafavre, line 101: Change the label so the appBar doesn't say "WishListViewFragment" but just "Wishlist"
```suggestion
        android:label="Wishlist" >
```

app/src/main/res/values/strings.xml
```diff
@@ -37,4 +40,6 @@
         <item>Clothes</item>
         <item>Miscellaneous</item>
     </string-array>
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
```
bafavre, line 44: What is that and when is it used?

bafavre (changes requested):
Nice work!
I have a few things that I think should be changed and also on my emulator, if I go to the detailed fragment without being logged in, the app crashes.
I don't know if it's just me, but it might be good to investigate this.

lumapools (approved):
Good job.

app/src/main/java/com/example/sharingang/DateConverter.kt
```diff
@@ -2,6 +2,7 @@ package com.example.sharingang
 
 import androidx.room.TypeConverter
 import java.util.*
+import kotlin.collections.ArrayList
```
bafavre, line 5: You can remove this.

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -13,16 +13,22 @@ import androidx.navigation.findNavController
...
+import javax.inject.Inject
 
 @AndroidEntryPoint
 class DetailedItemFragment : Fragment() {
+
     private val args: DetailedItemFragmentArgs by navArgs()
+    @Inject
+    lateinit var currentUserProvider: CurrentUserProvider
+    private val userViewModel : UserProfileViewModel by viewModels()
     private val viewModel: UserProfileViewModel by viewModels()
```
bafavre, line 32: Two times the same viewmodel?

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -31,6 +37,7 @@ class DetailedItemFragment : Fragment() {
         super.onCreate(savedInstanceState)
         observer = ImageAccess(requireActivity())
         lifecycle.addObserver(observer)
+        userViewModel.setUser(currentUserProvider.getCurrentUserId())
```
bafavre, line 40: Why do you set the user of the viewmodel here? We also set it elsewhere to correspond to the user that owns the item. Also since in your methods wishlistContains, modifyWishlist and refreshUi you call the currentUserProvider, you don't need to do it here.

bafavre (changes requested):
Still a few things to change, but almost done! :)

bafavre (approved):

PR #124, opened 2021-04-16
Fix bug that crashed the app when denying twice the permission request

### LINKED ISSUE #117: Fix Bug with permissions callbacks

v1ruso:
Fix #117 
Just a quick fix for the map feature: when denying the permission twice, it would slow down the app so much, making it unusable. The problem came from onResume, that was called every time the permission request window was shown, basically asking for permission endlessly.


lumapools (approved):
Well done for noticing where the issue came from. Good job!

cronokirby (approved):
Nice Catch!
image

PR #125, opened 2021-04-17
98 address translation

### LINKED ISSUE #98: Transform location (lat/long) into addresses

v1ruso:
Fix #98, #116 
I implemented a way to add addresses to items instead of geographic coordinates.

Screenshot 2021-04-17 at 20 19 33
Screenshot 2021-04-17 at 20 19 49
Screenshot 2021-04-17 at 20 20 04

A user can still use the button "Get Location", which will give an approximate address of the user, if possible (meaning that if the user is in the middle of the ocean, there won't be any address, but the coordinates will still be saved).

I also refactored the image icons and the design of new fragment, so that they take less space.

I didn't change EditFragment yet because I am planning on merging EditItemFragment and NewItemFragment in my next PR.

### DIFF
--- a/app/src/main/java/com/example/sharingang/NewItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/NewItemFragment.kt
@@ -165,6 +165,6 @@ class NewItemFragment : Fragment() {
         binding.longitude = location.longitude.toString()
         val address =
             geocoder.getFromLocation(location.latitude, location.longitude, 1).getOrNull(0)
-        binding.postalAddress.text = address?.getAddressLine(0)
+        binding.postalAddress.text = if(address!=null) address.getAddressLine(0) else ""
     }
 }
--- a/app/src/main/java/com/example/sharingang/NewItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/NewItemFragment.kt
@@ -2,8 +2,11 @@ package com.example.sharingang
 
 import android.net.Uri
 import android.Manifest
+import android.location.Address
+import android.location.Geocoder
 import android.location.Location
 import android.os.Bundle
+import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -11,6 +14,7 @@ import androidx.databinding.DataBindingUtil
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
 import androidx.navigation.findNavController
+import androidx.navigation.fragment.findNavController
 import com.example.sharingang.databinding.FragmentNewItemBinding
 import com.example.sharingang.items.Item
 import com.example.sharingang.items.ItemsViewModel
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/example/sharingang/NewItemFragment.kt
```diff
@@ -124,5 +163,8 @@ class NewItemFragment : Fragment() {
     private fun updateLocation(location: Location) {
         binding.latitude = location.latitude.toString()
         binding.longitude = location.longitude.toString()
+        val address =
+            geocoder.getFromLocation(location.latitude, location.longitude, 1).getOrNull(0)
+        binding.postalAddress.text = if(address!=null) address.getAddressLine(0) else ""
```
cronokirby, line 168:
```suggestion
        binding.postalAddress.text = address?.getAddressLine(0) ?: ""
```

I think this works, and is a bit more Kotlinesque

app/src/main/java/com/example/sharingang/NewItemFragment.kt
```diff
@@ -124,5 +163,8 @@ class NewItemFragment : Fragment() {
     private fun updateLocation(location: Location) {
         binding.latitude = location.latitude.toString()
         binding.longitude = location.longitude.toString()
+        val address =
+            geocoder.getFromLocation(location.latitude, location.longitude, 1).getOrNull(0)
```
cronokirby, line 167: This logic is repeated above, might be worth extracting into a method?

cronokirby (changes requested):
Just one minor null handling improvement, otherwise this looks pretty solid.

Good joberino!

app/src/main/java/com/example/sharingang/NewItemFragment.kt
v1ruso: Way cleaner, thanks!

app/src/main/java/com/example/sharingang/NewItemFragment.kt
v1ruso: Where is the logic repeated? If you're referring to `geocoder.getFromLocationName(place.address, 1).getOrNull(0)`, it's not the same: one goes from an address to a location, the other goes from a location to an address. 
Are you referring to the fact that the last argument is 1 and then we have `.getOrNull(0)`? If so, I don't know how to extract that any further, given how different the two methods are.

cronokirby: Oh you're right, it's `FromLocationName` instead, woops. Yeah, the methods look somewhat similar but are just opposites :P

v1ruso: I understand the confusion, no problem! :D

cronokirby (approved):
Looks good to me!

app/src/main/res/layout/fragment_new_item.xml
```diff
@@ -36,18 +38,27 @@
...
+                android:contentDescription="@string/item_image_description"
+                android:src="@drawable/ic_baseline_image_24" />
+
+            <ImageView
+                android:id="@+id/new_item_take_picture"
+                android:layout_width="match_parent"
+                android:layout_height="70dp"
+                android:layout_weight="1"
+                android:contentDescription="@string/item_image_description"
+                android:src="@android:drawable/ic_menu_camera" />
```
bafavre, line 60: Shouldn't we use a ImageButton with an ImageView here, since the image is not supposed to change?
Also I think having it on the right and the same size makes the image we add way too small, so maybe you should put it under it like before.

bafavre (changes requested):
Well done! :)
Just a small remark concerning the take picture image.

app/src/main/res/layout/fragment_new_item.xml
v1ruso: Indeed, I just copied what you already had for the other button.
As for the size, the previous version took too much space (since I'm adding a field the form has grown in size), so I needed to reduce the size of that. Improvements can be made, but this was the quickest fix.

v1ruso: Like having one button that opens two other buttons, to permits choosing between using the camera or using a picture from the gallery; instead of showing two buttons

v1ruso: I reverted it.

bafavre: I saw your changes, but now the create item button doesn't fit on my emulator (probably too small), can't we just make the whole fragment scrollable so we can always access everything?

v1ruso: I reduced the size of the form

bafavre: Alright, but at some point we'll have to make our fragments scrollable so it can be used on all screens. Maybe we can make some user stories for that.

bafavre (approved):

app/src/main/res/layout/fragment_new_item.xml
v1ruso: Agreed!

PR #130, opened 2021-04-20
Make map use searches

### LINKED ISSUE #121: Search on map
As a buyer, I want to be able to filter items in my surroundings according to certain keywords or categories.

### LINKED ISSUE #128: Allow filtering map items by category
See #121

cronokirby:
image
Fixes #128
Fixes #121 

This makes the map use the search results, instead of displaying all items. To make searching more convenient, I've also added a button to go from the map to the search fragment. This would be even more convenient if the map remembered your position, but that's for another story.

### DIFF
--- a/app/src/main/java/com/example/sharingang/MapFragment.kt
+++ b/app/src/main/java/com/example/sharingang/MapFragment.kt
@@ -39,12 +39,22 @@ class MapFragment : Fragment(), OnMapReadyCallback {
         )
     }
     private var lastLocation: Location? = null
-    private lateinit var locationCallback: LocationCallback
+    private val locationCallback: LocationCallback = object : LocationCallback() {
+        override fun onLocationResult(locationResult: LocationResult) {
+            lastLocation = locationResult.lastLocation
+            if (!hasCameraMovedOnce) {
+                moveCameraToLastLocation()
+                hasCameraMovedOnce = true
+            }
+            moveLastLocationMarker()
+        }
+    }
     private var lastLocationMarker: Marker? = null
     private var map: GoogleMap? = null
 
     private val viewModel: ItemsViewModel by activityViewModels()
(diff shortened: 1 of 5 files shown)


v1ruso (approved):
It works great, well done:)

app/src/main/res/layout/fragment_map.xml
```diff
@@ -6,20 +6,35 @@
...
+                app:icon="@android:drawable/ic_menu_mylocation" />
+
+            <Button
+                android:id="@+id/map_start_search"
+                style="@style/Widget.MaterialComponents.Button.Icon"
+                android:layout_width="55dp"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="10dp"
+                android:background="@drawable/rounded_corner"
+                app:icon="@android:drawable/ic_menu_search" />
```
v1ruso, line 37: I'd invert these two buttons: usually the "get location" button is the one in the corner

v1ruso (approved):

app/src/main/java/com/example/sharingang/MapFragment.kt
```diff
@@ -5,6 +5,7 @@ import android.annotation.SuppressLint
 import android.location.Location
 import android.os.Bundle
 import android.os.Looper
+import android.util.Log
```
bafavre, line 8: You can remove it if you don't use it I guess.

bafavre (approved):
Well done, only one nitpicking thing but otherwise it's all good! :)
It just feels kinda counter intuitive to have to search for items and then press the back button, but I can't think of something else as of now and with the user story related to it.

app/src/main/java/com/example/sharingang/MapFragment.kt
cronokirby: Good point!

Issue #122, opened 2021-04-16
Reporting Users


lumapools:
As a user, I would like to be able to report other users so that legitimate users will eventually not bump into many-times reported malicious ones.

PR #137, opened 2021-04-21
Report user

lumapools:
Fixes Issue #122 
Needs PR #138 to work.
A user has now the possibility of reporting a user. However, if the user is not logged in, they will not be able to report.

User profile view: non-logged-in user (left) versus logged-in user (right)
Webp net-resizeimage (5)   Webp net-resizeimage (6)

When the report is successful:
Webp net-resizeimage (7)


app/src/androidTest/java/com/example/sharingang/UserProfileFragmentTest.kt
```diff
@@ -64,6 +65,20 @@ class UserProfileFragmentTest {
         onView(withId(R.id.btnApply)).check(matches(not(isDisplayed())))
     }
 
+    @Test
+    fun reportUserIsDisplayedAndSendsReport() {
+        navigate_to(R.id.userProfileFragment)
+        onView(withId(R.id.btn_report)).check(matches(isDisplayed()))
+        onView(withId(R.id.btn_report)).perform(click())
+        val device: UiDevice = UiDevice.getInstance(getInstrumentation())
```
cronokirby, line 73:
```suggestion
        val device = UiDevice.getInstance(getInstrumentation())
```
Seems better to me without the explicit annotation, given the `UIDevice` appearing on the right as well. If this annotation is necessary, than ignore this

app/src/androidTest/java/com/example/sharingang/UserProfileFragmentTest.kt
```diff
@@ -64,6 +65,20 @@ class UserProfileFragmentTest {
...
+        onView(withText("Report Result")).check(matches(isDisplayed()));
+        onView(withText("Successfully reported Test User.")).check(matches(isDisplayed()));
+        val okButton = device.findObject(UiSelector().className("android.widget.Button"))
+        assert(okButton.text.toString() == "OK")
```
cronokirby, line 77: Shouldn't we use a `onView...check` here instead?

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -178,6 +183,31 @@ class UserProfileFragment : Fragment() {
...
+                val reporterUser = userRepository.get(currentUserId!!)!!
+                val reportedUser = userRepository.get(shownUserProfileId!!)!!
+                userRepository.report(reportedUser, reporterUser)
+            }
+            val builder = AlertDialog.Builder(requireContext())
+            builder.setTitle("Report Result")
+            builder.setMessage("Successfully reported $username.")
+            builder.setPositiveButton("OK") { _: DialogInterface, _: Int -> }
+            builder.show()
+
```
cronokirby, line 206:
```suggestion
```

(Note: I think you might've forgotten to format this file)

cronokirby (changes requested):
Just a few minor things, I think.

Otherwise, good job!

app/src/androidTest/java/com/example/sharingang/FakeCurrentUserProvider.kt
```diff
@@ -8,7 +8,7 @@ import kotlinx.coroutines.runBlocking
 import javax.inject.Inject
 
 class FakeCurrentUserProvider @Inject constructor(
-    private val userRepository: UserRepository
+    private val userRepository: UserRepository,
```
betrisey, line 11:
```suggestion
    private val userRepository: UserRepository
```

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
```diff
@@ -30,4 +35,22 @@ class FirestoreUserStore @Inject constructor(firestore: FirebaseFirestore) :
...
+    override suspend fun report(reportedUser: User, reporterUser: User): Boolean {
+        return try {
+            firestore
+                .collection("users").document(reportedUser.id!!)
+                .collection("reports").document(reporterUser.id!!).set(
+                    hashMapOf(
+                        reporterUser.id to Calendar.getInstance().time.toString()
+                    )
+                )
+            Log.d("TESTING FIRE", "User ${reporterUser.id} reported ${reportedUser.id}")
```
betrisey, line 47:
```suggestion
            Log.d(TAG, "User ${reporterUser.id} reported ${reportedUser.id}")
```

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
```diff
@@ -30,4 +35,22 @@ class FirestoreUserStore @Inject constructor(firestore: FirebaseFirestore) :
...
+            true
+        } catch (e: Exception) {
+            e.printStackTrace()
+            Log.e("TESTING FIRE", "Error reporting user ${reportedUser.id} by ${reporterUser.id}")
```
betrisey, line 51:
```suggestion
            Log.e(TAG, "Error reporting user ${reportedUser.id} by ${reporterUser.id}")
```

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
```diff
@@ -30,4 +35,22 @@ class FirestoreUserStore @Inject constructor(firestore: FirebaseFirestore) :
         return super.update(user, user.id)
     }
 
+    override suspend fun report(reportedUser: User, reporterUser: User): Boolean {
+        return try {
+            firestore
+                .collection("users").document(reportedUser.id!!)
+                .collection("reports").document(reporterUser.id!!).set(
+                    hashMapOf(
+                        reporterUser.id to Calendar.getInstance().time.toString()
```
betrisey, line 44: Will this create something like?
```json
{
  "reporter-id-4fe$f34lkd": "Apr 22 21:40:14"
}
```

If that's the case maybe it would be better to have
```json
{
  "reporter": "reporter-id-4fe$f34lkd",
  "reportedAt": "Apr 22 21:40:14"
}
```

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -178,6 +183,31 @@ class UserProfileFragment : Fragment() {
...
+                username = userRepository.get(shownUserProfileId!!)!!.name
+            }
+        }
+
+        binding.btnReport.setOnClickListener {
+            lifecycleScope.launch(Dispatchers.IO) {
+                val reporterUser = userRepository.get(currentUserId!!)!!
+                val reportedUser = userRepository.get(shownUserProfileId!!)!!
+                userRepository.report(reportedUser, reporterUser)
+            }
```
Superdoude1999, line 200: Why don't you create the function in the userViewModel rather than the user repository?

app/src/main/java/com/example/sharingang/users/InMemoryUserRepository.kt
```diff
@@ -42,4 +42,8 @@ class InMemoryUserRepository : UserRepository {
         usersMap[user.id] = user
         return true
     }
+
+    override suspend fun report(reportedUser: User, reporterUser: User): Boolean {
+        return false
```
betrisey, line 47: I think that even though we're doing nothing, we can return `true` so the UI will work correctly and display the success message

betrisey (changes requested):
Great!
Just check how it is stored in Firestore

Superdoude1999 (changes requested):
I approve of what you did, the only thing is you do a lot of logic inside the fragment itself, to the point where you call corroutines to the userRepository from it. Would it be possible to go through the userViewModel to do all of that computation in order to maintain the ViewModel pattern. Basically it works, it's fine, but it doesn't follow the pattern is what bothers me.

app/src/androidTest/java/com/example/sharingang/FakeCurrentUserProvider.kt
lumapools: I'm not sure I understand this suggestion

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
lumapools: Oh my I completely forgot about that! Thanks!

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
lumapools: Okay, I will try!

app/src/main/java/com/example/sharingang/users/InMemoryUserRepository.kt
lumapools: Got it!

app/src/androidTest/java/com/example/sharingang/FakeCurrentUserProvider.kt
betrisey: I think that there is an unnecessary comma (maybe I don't see it correctly on the review page)

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
betrisey: And I think that you can use `Date()` without converting it into a String.
Firestore can store it as a date

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
lumapools: So you prefer something like this:
```
fun reportUser(reportedUser: User, reporterUser: User)  {
            val reporterUser = userRepository.get(currentUserId!!)!!
            val reportedUser = userRepository.get(shownUserProfileId!!)!!
            userRepository.report(reportedUser, reporterUser)
    }
}
```
inside UserViewModel and then call `viewModel.report(reportedUser, reporterUser)` from the UserProfileFragment?

Superdoude1999: Yes that's essentially what I mean

betrisey (approved):

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -162,22 +179,44 @@ class UserProfileFragment : Fragment() {
...
+            lifecycleScope.launch(Dispatchers.IO) {
+                username = userRepository.get(shownUserProfileId!!)!!.name
+            }
+        }
+
+        binding.btnReport.setOnClickListener {
+            userViewModel.report(shownUserProfileId!!, currentUserId!!)
+            val builder = AlertDialog.Builder(requireContext())
+            builder.setTitle("Report Result")
+            builder.setMessage("Successfully reported $username.")
```
Superdoude1999, line 214: Last thing and I'm sorry, but can you make it so the report function returns the username that way you don't have to user another coroutine in the fragment to get it. Just get it in the viewModel while you report and then return the username. I can take care of that another day if you want though.

app/src/main/java/com/example/sharingang/UserProfileViewModel.kt
```diff
@@ -87,5 +87,15 @@ class UserProfileViewModel @Inject constructor(
...
     }
+
+    fun report(reportedUserId: String, reporterUserId: String)  {
+        viewModelScope.launch(Dispatchers.IO) {
+            val reporterUser = userRepository.get(reporterUserId)
+            val reportedUser = userRepository.get(reportedUserId)
+            userRepository.report(reportedUser!!, reporterUser!!)
+        }
+    }
+
```
Superdoude1999, line 98: very nice ;)

Superdoude1999 (approved):

cronokirby (approved):
Good job improving on all of the other wonderful comments!

PR #138, opened 2021-04-21
Fix bug for logged-out users navigating to user profile.

lumapools:
Fix: Now the user can see other users' profile without being logged in.


bafavre (approved):
All good!
Just a question, should we be bale to see the email address of other users on their profile or not?

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -112,17 +112,15 @@ class UserProfileFragment : Fragment() {
 
     private fun setEmailText() {
         val emailText = binding.textEmail
-        if(isAuthUserDisplayedUser() && currentUserId != null) {
+        if(currentUserId != null && isAuthUserDisplayedUser()) {
             emailText.text = loggedInUserEmail
-        }
-        if(currentUserId != null) {
             emailText.visibility = View.VISIBLE
```
v1ruso, line 117: I agree with @bafavre's comment: is the email visible to all the other users?

app/src/main/res/values/strings.xml
```diff
@@ -36,7 +36,7 @@
     <string name="remove_wishlist">Remove wishlist</string>
     <string name="wishlist_menu_title">Wish List</string>
     <string name="enter_address">Enter Address</string>
-    <string name="upfNotLoggedInInfo">You need to be logged in to view User Profiles.</string>
+    <string name="upfNotLoggedInInfo">You need to be logged in to view your User Profile.</string>
```
v1ruso, line 39: "upfNot...", I'd rename this to `userNotLoggedInInfo`or similar

v1ruso (changes requested):

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
v1ruso: Indeed, which is why I wanted to make sure

v1ruso (approved):
Nice

PR #140, opened 2021-04-22
126 merge edit new

v1ruso:
#126 
Nothing new really, just that NewItem and EditItem fragments don't exist anymore, to be replaced with NewEditFragment.

To goal was to avoid copy pasted code between the two fragments.
Screenshot 2021-04-22 at 12 33 48
Screenshot 2021-04-22 at 12 34 14


app/src/main/java/com/example/sharingang/NewEditFragment.kt
```diff
@@ -161,10 +174,45 @@ class NewItemFragment : Fragment() {
...
+        binding.latitude = latitude.toString()
+        binding.longitude = longitude.toString()
         val address =
-            geocoder.getFromLocation(location.latitude, location.longitude, 1).getOrNull(0)
+            geocoder.getFromLocation(latitude, longitude, 1).getOrNull(0)
         binding.postalAddress.text = address?.getAddressLine(0) ?: ""
     }
+
+    private fun setupNewOrEditFragment() {
+        if (viewModel.onEdit.value == false) {
```
betrisey, line 189: Can we do `existingItem = args.item` before and check if it's null to determine if we are creating a new item or editing an existing item?

betrisey (approved):
Nice!
Just a remark, the onEdit in the viewModel add some complexity. Is it possible to remove it and just check if the item passed as argument to the fragment is null?

app/src/main/java/com/example/sharingang/NewEditFragment.kt
```diff
@@ -161,10 +174,45 @@ class NewItemFragment : Fragment() {
...
+            binding.editItemPrompt.visibility = View.GONE
+            binding.editItemButton.visibility = View.GONE
+            binding.newItemPrompt.visibility = View.VISIBLE
+            binding.createItemButton.visibility = View.VISIBLE
+        } else {
+            binding.editItemPrompt.visibility = View.VISIBLE
+            binding.editItemButton.visibility = View.VISIBLE
+            binding.newItemPrompt.visibility = View.GONE
+            binding.createItemButton.visibility = View.GONE
```
bafavre, line 199: Maybe it would be easier to work with only one prompt and one button and change their appearance/behavior depending on whether an item is passed as an argument. Instead of having two of each, but I don't know if it's easier and clearer.

bafavre (commented):
Good job!
I agree on the remark Sam made and I also have a few of my own that you can check.

app/src/main/java/com/example/sharingang/NewEditFragment.kt
```diff
@@ -111,41 +115,50 @@ class NewItemFragment : Fragment() {
...
             category = binding.categorySpinner.selectedItemPosition,
             categoryString = resources.getStringArray(R.array.categories)[binding.categorySpinner.selectedItemPosition],
             latitude = binding.latitude?.toDoubleOrNull() ?: 0.0,
             longitude = binding.longitude?.toDoubleOrNull() ?: 0.0,
-            sold = false,
-            imageUri = imageUri?.toString(),
-            userId = userId
+            userId = existingItem?.userId ?: userId,
+            createdAt = existingItem?.createdAt,
+            localId = existingItem?.localId ?: 0
```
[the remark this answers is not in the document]
v1ruso: This value of 0 should be ignored, because in Item class the localId is autogenerated.
By removing "0", it fails because localId cannot be null.

app/src/main/java/com/example/sharingang/NewEditFragment.kt
v1ruso: By doing so we would need to change their appearance and their action. I thought it was easier this way. Also, if we want to change the GUI later one, we can do it directly in the XML file, without going through the fragment.

app/src/main/java/com/example/sharingang/NewEditFragment.kt
v1ruso: Corrected, thanks!

Superdoude1999 (approved):

bafavre (approved):

PR #144, opened 2021-04-22
Detailed markers on the map

betrisey:
I replaced the marker with a box that contains the title and the price of the item.
When the items are too close, they are clustered together.
We can easily customize this box, it's an xml layout.

https://user-images.githubusercontent.com/1939395/115712965-be0dab00-a375-11eb-966b-45ff23d2185d.mp4


app/src/main/java/com/example/sharingang/MapFragment.kt
```diff
@@ -62,41 +68,42 @@ class MapFragment : Fragment(), OnMapReadyCallback {
...
+        clusterManager?.cluster()
     }
 
     @SuppressLint("MissingPermission")
     private fun startLocationUpdates() {
+        // Show blue dot at current location
+        map?.isMyLocationEnabled = true
+        // We already have a custom button so we remove the default one
+        map?.uiSettings?.isMyLocationButtonEnabled = false
+
```
v1ruso, line 106: What happens if the location permission is not granted?

app/src/main/java/com/example/sharingang/MapFragment.kt
```diff
@@ -45,11 +50,12 @@ class MapFragment : Fragment(), OnMapReadyCallback {
...
                 hasCameraMovedOnce = true
             }
-            moveLastLocationMarker()
         }
     }
-    private var lastLocationMarker: Marker? = null
+
     private var map: GoogleMap? = null
+    private var markerManager: MarkerManager? = null
+    private var clusterManager: ClusterManager<MapItem>? = null
```
v1ruso, line 58: Can we make those lateinit vars? To avoid having unnecessary null checks

app/src/main/java/com/example/sharingang/MapFragment.kt
```diff
@@ -45,11 +50,12 @@ class MapFragment : Fragment(), OnMapReadyCallback {
                 moveCameraToLastLocation()
                 hasCameraMovedOnce = true
             }
-            moveLastLocationMarker()
         }
     }
-    private var lastLocationMarker: Marker? = null
+
     private var map: GoogleMap? = null
```
v1ruso, line 56: Same as below comment

app/src/main/java/com/example/sharingang/MapFragment.kt
```diff
@@ -120,42 +126,36 @@ class MapFragment : Fragment(), OnMapReadyCallback {
...
         }
-        addItemMarkers(viewModel.searchResults.value ?: listOf())
 
-        doOrGetPermission(
-            this,
-            Manifest.permission.ACCESS_FINE_LOCATION,
-            { startLocationUpdates() },
-            requestPermissionLauncher
-        )
+        map!!.setOnCameraIdleListener(clusterManager)
```
v1ruso, line 158: What does this do?

app/src/main/res/layout/item_marker.xml
```diff
@@ -0,0 +1,22 @@
...
+    android:layout_height="match_parent"
+    android:orientation="vertical">
+
+    <TextView
+        android:id="@+id/titleText"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:textAlignment="center"
+        android:textSize="18sp"
+        tools:text="Naruto book" />
```
v1ruso, line 14: What is that?

app/src/main/res/layout/item_marker.xml
```diff
@@ -0,0 +1,22 @@
...
+
+    <TextView
+        android:id="@+id/descriptionText"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:textAlignment="center"
+        tools:text="$ 42.00" />
```
v1ruso, line 21: Same question?

app/src/main/java/com/example/sharingang/MapFragment.kt
betrisey: I think that this function is only called when the permission has been granted.
I'm not sure, could you check that please? This function already existed before, I only added these 2 lines

app/src/main/java/com/example/sharingang/MapFragment.kt
betrisey: I really wanted to do that too but I think it's not possible because it requires the map and it's not initialized in `onCreateView`, it's done in a coroutine because we have to do it asynchronously

betrisey: A way to avoid all these null checks would be to pass around the `map` object to the functions that need it but I don't know if it's better. What do you think ?

app/src/main/java/com/example/sharingang/MapFragment.kt
betrisey: `OnCameraIdle` is the event when the user finished moving the map. The cluster manager needs to receive this event so it can recompute the cluster (for example when the user zoom out, it would cluster them together)

app/src/main/res/layout/item_marker.xml
betrisey: This is the XML layout of the marker used in the map
Screenshot 2021-04-22 at 14 59 35

app/src/main/java/com/example/sharingang/MapFragment.kt
v1ruso: Oh okk my bad, then all good!

app/src/main/java/com/example/sharingang/MapFragment.kt
v1ruso: You're right, then I guess it's fine the way it is

app/src/main/java/com/example/sharingang/MapFragment.kt
v1ruso: Thanks!

app/src/main/res/layout/item_marker.xml
v1ruso: Yes but why the "naruto book"?

Superdoude1999 (approved):

app/src/main/res/layout/item_marker.xml
betrisey: every attribute that begins with `tools:` is only visible in Android Studio when creating the layout
So it's just a placeholder value to avoid having blank TextViews 
Screenshot 2021-04-22 at 15 24 56

v1ruso (approved):
Well done

app/src/main/res/layout/item_marker.xml
bafavre: This is a tools:text, which is a design time text, it doesn't show up in the app.

PR #146, opened 2021-04-22
Wish list visibility

Superdoude1999:
Changed the wish list button to only be visible if logged in.
When going on the actual wish list fragment, a toast pops up to indicate if you are not logged in and hence that you can't see your wish list.


betrisey (approved):

v1ruso (approved):
Nice

PR #152, opened 2021-04-22
Save images on Firebase

betrisey:
I already implemented the saving on Firebase.
@Superdoude1999 will then update the UI to use it


app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
```diff
@@ -72,7 +76,12 @@ class ItemsViewModel @Inject constructor(
      */
     fun setItem(item: Item, callback: ((String?) -> Unit)? = null) {
         viewModelScope.launch(Dispatchers.IO) {
-            val itemId = itemRepository.set(item)
+            val uploadUrl = item.image?.let {
+                if(!it.startsWith("https://")){
+                    imageStore.store(it.toUri())
+                }
+            }
+            val itemId = itemRepository.set(item.copy(image = uploadUrl.toString()))
```
betrisey, line 84: You have to add an else where you return the current URL so you don't lose the existing image
```suggestion
            val uploadUrl = item.image?.let {
                if (!it.startsWith("https://")) {
                    imageStore.store(it.toUri()).toString()
                } else {
                    it
                }
            }
            val itemId = itemRepository.set(item.copy(image = uploadUrl))
```

Superdoude1999: Yup thank you for the catch, modified and implemented in the latest commit

Superdoude1999 (approved):
Sneaky reviews

Superdoude1999 (approved):
thank you sir

bafavre (approved):
All good, well done!

PR #153, opened 2021-04-22
119 rating view

Superdoude1999:
Added possibility to rate an item if the user is logged in, the object hasn't been rated yet (to be changed so that only the buyer can rate it) and the object has been sold. Rating is stored in the user object and then on the database. If a user has no ratings yet, a simple "No Ratings" is displayed instead of his rating.

This is when an item is not yet sold.

This is when an item is sold and you are signed in, the rating views appear and you can rate the object. Once you hit the rate button it looks like the first picture again and no one who accesses the item from then on will have the option to rate it.

This is what the user profile looks like with no ratings.

And finally this is what a user profile with a few reviews looks like.


app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -52,6 +54,7 @@ class DetailedItemFragment : Fragment() {
         }
 
         initiateWishlistButton(binding)
+        initRating(binding)
```
betrisey, line 57: Maybe better to make binding a `private lateinit var` attribute of the class because you need it in a few methods

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -80,7 +83,34 @@ class DetailedItemFragment : Fragment() {
         }else{
             binding.addToWishlist.visibility = View.GONE;
         }
+    }
+
+    private fun initRating(binding: FragmentDetailedItemBinding){
+        updateRatingVisibility(binding)
+        binding.ratingButton.setOnClickListener {
+            val selectedOPtion: Int = binding.radioGroup1.checkedRadioButtonId
```
betrisey, line 91: small typo, `P` should be lowercase

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -83,13 +85,25 @@ class UserProfileFragment : Fragment() {
...
         )
         for (view: View in fields) {
             view.visibility = View.GONE
         }
     }
 
+    private fun setupRatingView(){
+        userViewModel.refreshRating(shownUserProfileId)
+        userViewModel.rating.observe(viewLifecycleOwner, {
+            var text = resources.getString(R.string.default_rating)
```
betrisey, line 100: This variable looks unused

lumapools (commented):
Good work! I would just think it might be more understandable to have the "out of" shown. It makes sense that 5.00 is the maximum rating level as it is a commonly used number for this, but I think having `rating/5.00` would be more "user-friendly". However, it might be too much "useless" information, what are your thoughts on it?

Also I am not sure I fully understand how rating works. When does it happen? Could you provide some pictures of the interface please?

app/src/main/java/com/example/sharingang/items/Item.kt
```diff
@@ -46,6 +46,8 @@ data class Item(
     /** Id of the user that created the item */
     val userId: String? = null,
 
+    var rated: Boolean = false,
```
betrisey, line 49: If possible, try to have keep the fields of Item and User immutable.
You can easily make a copy of an item and change a simple field:
`item.copy(rated = true)`

app/src/main/java/com/example/sharingang/users/User.kt
```diff
@@ -29,6 +29,9 @@ data class User(
     /** URLs of the images */
     val profilePicture: String? = null,
 
+    var rating: Int = 0,
+    var numberOfRatings: Int = 0,
+
     var wishlist: List<String> = ArrayList(),
```
betrisey, line 35: Would be better to make it immutable too

betrisey (changes requested):
Good job overall!

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
Superdoude1999: good call

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
Superdoude1999: oops was meant to be set to ratingTextview

app/src/main/java/com/example/sharingang/users/User.kt
Superdoude1999: and done

Superdoude1999:
> Good work! I would just think it might be more understandable to have the "out of" shown. It makes sense that 5.00 is the maximum rating level as it is a commonly used number for this, but I think having `rating/5.00` would be more "user-friendly". However, it might be too much "useless" information, what are your thoughts on it?
> 
> Also I am not sure I fully understand how rating works. When does it happen? Could you provide some pictures of the interface please?

Hey so I'll maybe add the /5.00 later if you guys think it's a good idea but I feel like it's so self explanatory that it doesn't really need it. And as per your request I have provided some nice little photos in the description

betrisey (approved):
GG

lumapools (approved):
Very nice!

PR #166, opened 2021-04-24
Improve report user

lumapools:
Report user has now its own view. There are categories from which the reporter can choose from (exactly one choice), as well as provide a description of why a particular user is being reported. A user can report another user only once, and they also cannot report themselves.

Webp net-resizeimage (10) Webp net-resizeimage (9)

The report is put into the database, where the reason, the reporter + reported user id, the description and the date are recorded.
Note: The OK button is not clickable unless the reporter has not clicked a category / reason. The description is optional.

### DIFF
--- a/app/src/androidTest/java/com/example/sharingang/ReportFragmentTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/ReportFragmentTest.kt
@@ -2,6 +2,7 @@ package com.example.sharingang
 
 
 import android.Manifest
+import androidx.test.espresso.Espresso
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions
 import androidx.test.espresso.action.ViewActions.*
@@ -29,11 +30,21 @@ class ReportFragmentTest {
 
     @Test
     fun reportCanBeCancelled() {
-        navigate_to(R.id.userProfileFragment)
+        FakeCurrentUserProvider.instance = 2
+        navigate_to(R.id.newEditFragment)
+        Thread.sleep(500)
+        onView(withId(R.id.itemTitle)).perform(
+            typeText("TestItem"),
+            closeSoftKeyboard()
+        )
+        onView(withId(R.id.createItemButton)).perform(click())
+        FakeCurrentUserProvider.instance = 1
+        onView(withId(R.id.item_list_view_title)).perform(click())
+        onView(withId(R.id.itemPostedBy)).perform(click())
         Thread.sleep(500)
         onView(withId(R.id.btn_report)).check(matches(isDisplayed()))
         onView(withId(R.id.btn_report)).perform(click())
-        onView(withId(R.id.text_reportedUsername)).check(matches(withText("Reporting Test User")))
+        onView(withId(R.id.text_reportedUsername)).check(matches(withText("Reporting Test User 2")))
         onView(withId(R.id.button_ok)).check(matches(not(isEnabled())))
         onView(withId(R.id.radio_item)).perform(click())
         onView(withId(R.id.button_ok)).check(matches(isEnabled()))
@@ -44,11 +55,30 @@ class ReportFragmentTest {
 
     @Test
     fun reportCanBeSent() {
-        navigate_to(R.id.userProfileFragment)
+        FakeCurrentUserProvider.instance = 2
+        navigate_to(R.id.newEditFragment)
         Thread.sleep(500)
+        onView(withId(R.id.itemTitle)).perform(
+            typeText("TestItem"),
+            closeSoftKeyboard()
+        )
+        onView(withId(R.id.createItemButton)).perform(click())
+        FakeCurrentUserProvider.instance = 1
+        onView(withId(R.id.item_list_view_title)).perform(click())
+        onView(withId(R.id.itemPostedBy)).perform(click())
+        Thread.sleep(500)
+        onView(withId(R.id.btn_report)).check(matches(isDisplayed()))
         onView(withId(R.id.btn_report)).perform(click())
+        onView(withId(R.id.text_reportedUsername)).check(matches(withText("Reporting Test User 2")))
         onView(withId(R.id.radio_other)).perform(click())
         onView(withId(R.id.button_ok)).perform(click())
+    }
+
+    @Test
+    fun aUserCannotReportThemselves() {
+        FakeCurrentUserProvider.instance = 1
+        navigate_to(R.id.userProfileFragment)
+        onView(withId(R.id.btn_report)).check(matches(not(isDisplayed())))
     }
 
 
--- a/app/src/androidTest/java/com/example/sharingang/ReportFragmentTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/ReportFragmentTest.kt
@@ -1,0 +1,55 @@
+package com.example.sharingang
+
+
+import android.Manifest
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.action.ViewActions.*
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.*
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
+import androidx.test.rule.GrantPermissionRule
+import androidx.test.uiautomator.UiDevice
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.hamcrest.CoreMatchers.not
+import org.junit.Rule
+import org.junit.Test
+
+@HiltAndroidTest
+class ReportFragmentTest {
+
+    @get:Rule(order = 0)
+    val hiltRule = HiltAndroidRule(this)
+
+    // We start with the main activity, and then navigate where we want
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -0,0 +1,107 @@
...
+        reportedUsername = args.reportedName
+
+
+        val radioButtons = listOf(
+            binding.radioUsername, binding.radioProfilePicture, binding.radioItem, binding.radioOther
+        )
+        setupRadioButtons(radioButtons)
+        setupCancelButton()
+        setupButtonOk(radioButtons)
+        binding.textReportedUsername.text = "Reporting $reportedUsername"
```
v1ruso, line 47: I'd add a string in values/strings.xml for that. Also check the warning: "don't concatenate text"

app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private fun setupRadioButtons(radioButtons: List<RadioButton>) {
+        for(radioButton: RadioButton in radioButtons) {
+            radioButton.setOnCheckedChangeListener { _, _ ->
+                binding.buttonOk.isEnabled =
+                    binding.radioUsername.isChecked || binding.radioProfilePicture.isChecked ||
+                    binding.radioItem.isChecked || binding.radioOther.isChecked
+            }
+        }
```
v1ruso, line 59: Isn't there a way to use a function from RadioGroup instead of making a loop?

app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private fun setupButtonOk(radioButtons: List<RadioButton>) {
+        binding.buttonOk.setOnClickListener { view ->
+            var reason = ""
+            for (radioButton: RadioButton in radioButtons) {
+                if (radioButton.isChecked) {
+                    reason = getName(radioButton)
+                }
+            }
```
v1ruso, line 77: Here again, isn't there a function in RadioGroup to get the selected button?

app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -0,0 +1,107 @@
...
+            )
+        }
+    }
+
+    private fun getName(radioButton: RadioButton): String {
+        return when (radioButton) {
+            binding.radioUsername -> "Inappropriate Username"
+            binding.radioProfilePicture -> "Inappropriate Profile Picture"
+            binding.radioItem -> "Inappropriate Item"
+            binding.radioOther -> "Other"
```
v1ruso, line 102: I'd add strings in xml/strings.xml, in case we, for some reason, decide to change the language for example

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
```diff
@@ -220,23 +226,23 @@ class UserProfileFragment : Fragment() {
...
     private fun setupReportButton() {
-        var username = ""
-        if (currentUserId != null) {
-            binding.btnReport.visibility = View.VISIBLE
-            lifecycleScope.launch(Dispatchers.IO) {
-                username = userRepository.get(shownUserProfileId!!)!!.name
+        if(currentUserId != null) {
+            lifecycleScope.launch(Dispatchers.Main) {
+                val hasBeenReported =
+                    userRepository.hasBeenReported(currentUserId!!, shownUserProfileId!!)
```
v1ruso, line 232: A user can report him/her-self, I don't think one should be able to do that

v1ruso (changes requested):

app/src/main/java/com/example/sharingang/UserProfileFragment.kt
lumapools: For now I don't know what else I can do, because I have not found a method for logging in two different users during testing, so this would probably give me ~20% coverage overall. Any suggestions?

app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -37,25 +37,17 @@ class ReportFragment : Fragment() {
...
-        val radioButtons = listOf(
-            binding.radioUsername, binding.radioProfilePicture, binding.radioItem, binding.radioOther
-        )
-        setupRadioButtons(radioButtons)
+        setupRadioButtons()
         setupCancelButton()
-        setupButtonOk(radioButtons)
-        binding.textReportedUsername.text = "Reporting $reportedUsername"
+        setupButtonOk()
+        "Reporting $reportedUsername".also { binding.textReportedUsername.text = it }
```
v1ruso, line 43: Isn't there a way not to have the string hardcoded?

app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -37,25 +37,17 @@ class ReportFragment : Fragment() {
...
-    private fun setupRadioButtons(radioButtons: List<RadioButton>) {
-        for(radioButton: RadioButton in radioButtons) {
-            radioButton.setOnCheckedChangeListener { _, _ ->
-                binding.buttonOk.isEnabled =
-                    binding.radioUsername.isChecked || binding.radioProfilePicture.isChecked ||
-                    binding.radioItem.isChecked || binding.radioOther.isChecked
-            }
+    private fun setupRadioButtons() {
+        binding.reportRadioGroup.setOnCheckedChangeListener { _, _ ->
+            binding.buttonOk.isEnabled = true
```
v1ruso, line 50: Well done

app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -67,14 +59,11 @@ class ReportFragment : Fragment() {
...
+    private fun setupButtonOk() {
         binding.buttonOk.setOnClickListener { view ->
-            var reason = ""
-            for (radioButton: RadioButton in radioButtons) {
-                if (radioButton.isChecked) {
-                    reason = getName(radioButton)
-                }
-            }
+            val checkedGroup = binding.reportRadioGroup.checkedRadioButtonId
+            val reason = getName(checkedGroup)
```
v1ruso, line 65: Nice!

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
```diff
@@ -63,6 +63,6 @@ class FirestoreUserStore @Inject constructor(private val firestore: FirebaseFire
     override suspend fun hasBeenReported(reporterId: String, reportedId: String): Boolean {
         val docIdRef = firestore.collection("users").document(reportedId)
             .collection("reports").document(reporterId).get().await()
-        return docIdRef.exists()
+        return docIdRef.exists() || reportedId == reporterId
```
v1ruso, line 66: If the reported id equals the id of the reporter return true? Shouldn't that be an "and"? Maybe I just misunderstood the purpose of this function

v1ruso (approved):

app/src/main/java/com/example/sharingang/users/FirestoreUserStore.kt
lumapools: TL;DR: Actually this change is something I wanted to remove from my last commit because the "or" check is useless.

Its purpose is to return whether the reported has already been reported by the reporter user, therefore if it returns true, then it means that the reporter user does not have the right to report the reported user. In this commit, it also returns true when the user wants to report themself (therefore later on not allowing this action). However, as I have actually moved this check to UserProfileFragment for better understanding, this check becomes useless, and I will remove it.

app/src/main/java/com/example/sharingang/ReportFragment.kt
lumapools: I could move "Reporting" to strings.xml, but then I would still need to concatenate, so I'm not really sure about it. Do you prefer that I make a string resource for "Reporting" and then set the text to `getString(...) $reportedUsername`?

app/src/androidTest/java/com/example/sharingang/ReportFragmentTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    @get:Rule(order = 0)
+    val hiltRule = HiltAndroidRule(this)
+
+    // We start with the main activity, and then navigate where we want
+    @get:Rule(order = 1)
+    val activityRule = ActivityScenarioRule(MainActivity::class.java)
+
+    @Test
+    fun reportCanBeCancelled() {
+        FakeCurrentUserProvider.instance = 2
```
betrisey, line 33: It's a clever way to do it, nice :)

app/src/main/java/com/example/sharingang/ReportFragment.kt
```diff
@@ -0,0 +1,96 @@
...
+            }
+            view.findNavController().navigate(
+                ReportFragmentDirections.actionReportFragmentToUserProfileFragment(
+                    reportedId
+                )
+            )
+        }
+    }
+
+    private fun getName(radioButtonId: Int): String {
```
betrisey, line 87:
```suggestion
    private fun getReasonName(radioButtonId: Int): String {
```

betrisey (changes requested):
Great job :)
Just check again if this is possible to do the database call on the IO thread otherwise everything's good!

betrisey (approved):
Thanks

PR #173, opened 2021-04-28
Edit and Sell in AppBar

### LINKED ISSUE #156: Edit and Sell from Item Detail
As a Seller, I want to have a centralized view per item where I can edit and mark the item as sold.

### LINKED ISSUE #167: Move Edit Button in Detailed Item Fragment

### LINKED ISSUE #168: Move Sell Button in Top Right Menu in Detailed Item Fragment

bafavre:
Fixes #167 
Fixes #168 
Fixes #156 

Now the ugly buttons from the item list have been removed and replaced by Icons on the AppBar in the detailed view fragment.

### DIFF
--- a/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
@@ -3,6 +3,7 @@ package com.example.sharingang
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions
 import androidx.test.espresso.assertion.ViewAssertions
+import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.matcher.ViewMatchers
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
@@ -38,10 +39,10 @@ class SellItemFragmentTest {
         buttonCreate.check(ViewAssertions.matches(ViewMatchers.withText("Create Item")))
         buttonCreate.perform(ViewActions.click())
 
-        onView(ViewMatchers.withText("Sell")).perform(ViewActions.click())
-        // Wait for RecyclerView to remove old button
-        Thread.sleep(1000)
-        onView(ViewMatchers.withId(R.id.item_list_view_sell_button))
-            .check(ViewAssertions.matches(ViewMatchers.withText("Resell")))
+        onView(ViewMatchers.withId(R.id.item_list_view_title)).perform(ViewActions.click())
+
+        onView(ViewMatchers.withId(R.id.menuSell)).perform(ViewActions.click())
+
+        onView(ViewMatchers.withId(R.id.menuResell)).check(matches(ViewMatchers.isDisplayed()))
     }
 }
--- a/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
@@ -2,9 +2,8 @@ package com.example.sharingang
 
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions
-import androidx.test.espresso.assertion.ViewAssertions
-import androidx.test.espresso.assertion.ViewAssertions.matches
-import androidx.test.espresso.matcher.ViewMatchers
+import androidx.test.espresso.assertion.ViewAssertions.*
+import androidx.test.espresso.matcher.ViewMatchers.*
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import dagger.hilt.android.testing.HiltAndroidRule
@@ -29,20 +28,20 @@ class SellItemFragmentTest {
     @Test
     fun anItemCanBeEditedAndSeenOnItemsListFragment() {
         navigate_to(R.id.newEditFragment)
-        onView(ViewMatchers.withId(R.id.newItemPrompt))
-            .check(ViewAssertions.matches(ViewMatchers.withText("New Item")))
-        onView(ViewMatchers.withId(R.id.itemTitle)).perform(
+        onView(withId(R.id.newItemPrompt))
+            .check(matches(withText("New Item")))
+        onView(withId(R.id.itemTitle)).perform(
             ViewActions.typeText(itemTitle),
             ViewActions.closeSoftKeyboard()
         )
-        val buttonCreate = onView(ViewMatchers.withId(R.id.createItemButton))
-        buttonCreate.check(ViewAssertions.matches(ViewMatchers.withText("Create Item")))
+        val buttonCreate = onView(withId(R.id.createItemButton))
+        buttonCreate.check(matches(withText("Create Item")))
         buttonCreate.perform(ViewActions.click())
 
-        onView(ViewMatchers.withId(R.id.item_list_view_title)).perform(ViewActions.click())
+        onView(withId(R.id.item_list_view_title)).perform(ViewActions.click())
 
-        onView(ViewMatchers.withId(R.id.menuSell)).perform(ViewActions.click())
+        onView(withId(R.id.menuSell)).perform(ViewActions.click())
 
-        onView(ViewMatchers.withId(R.id.menuResell)).check(matches(ViewMatchers.isDisplayed()))
+        onView(withId(R.id.menuResell)).check(matches(isDisplayed()))
     }
 }
--- a/app/src/androidTest/java/com/example/sharingang/WishlistViewTest.kt
+++ b/app/src/androidTest/java/com/example/sharingang/WishlistViewTest.kt
@@ -52,7 +52,7 @@ class WishlistViewTest {
         onView(withId(R.id.addToWishlist)).check(matches(withText(R.string.remove_wishlist)))
         onView(withId(R.id.addToWishlist)).perform(click())
         onView(isRoot()).perform(pressBack())
-
+        Thread.sleep(1000)
         onView(withText(secondItem)).check(matches(isDisplayed()))
     }
 
--- a/app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
@@ -157,7 +157,6 @@ class DetailedItemFragment : Fragment() {
             else View.GONE
             binding.ratingVisibility = visibility
         })
-
         binding.ratingButton.setOnClickListener {
             val selectedOption: Int = binding.radioGroup1.checkedRadioButtonId
             if (selectedOption != -1) {
--- a/app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
@@ -3,9 +3,7 @@ package com.example.sharingang
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
-import android.view.LayoutInflater
-import android.view.View
-import android.view.ViewGroup
+import android.view.*
 import androidx.databinding.DataBindingUtil
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
@@ -32,11 +30,13 @@ class DetailedItemFragment : Fragment() {
     private val viewModel: UserProfileViewModel by viewModels()
     private val itemViewModel: ItemsViewModel by viewModels()
     private lateinit var binding: FragmentDetailedItemBinding
+    private var soldStatus: Boolean = false
 
     private lateinit var observer: ImageAccess
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        setHasOptionsMenu(true)
         observer = ImageAccess(requireActivity())
         lifecycle.addObserver(observer)
     }
@@ -46,7 +46,7 @@ class DetailedItemFragment : Fragment() {
         savedInstanceState: Bundle?
     ): View {
         binding = DataBindingUtil.inflate(inflater, R.layout.fragment_detailed_item, container, false)
-
+        soldStatus = args.item.sold
         binding.item = args.item
         observer.setupImageView(binding.detailedItemImage)
         args.item.imageUri?.let {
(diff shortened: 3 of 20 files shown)


app/src/androidTest/java/com/example/sharingang/SellItemFragmentTest.kt
```diff
@@ -38,10 +39,10 @@ class SellItemFragmentTest {
...
-        onView(ViewMatchers.withText("Sell")).perform(ViewActions.click())
-        // Wait for RecyclerView to remove old button
-        Thread.sleep(1000)
-        onView(ViewMatchers.withId(R.id.item_list_view_sell_button))
-            .check(ViewAssertions.matches(ViewMatchers.withText("Resell")))
+        onView(ViewMatchers.withId(R.id.item_list_view_title)).perform(ViewActions.click())
+
+        onView(ViewMatchers.withId(R.id.menuSell)).perform(ViewActions.click())
+
+        onView(ViewMatchers.withId(R.id.menuResell)).check(matches(ViewMatchers.isDisplayed()))
```
cronokirby, line 46: We should probably change this code to either always use qualified imports (e.g. `ViewAssertions.matches`) or to just import `ViewAssertions.*`, and use `matches`. (I'd prefer the latter, but it doesn't really matter; consistency is more important).

app/src/androidTest/java/com/example/sharingang/WishlistViewTest.kt
```diff
@@ -52,7 +52,7 @@ class WishlistViewTest {
         onView(withId(R.id.addToWishlist)).check(matches(withText(R.string.remove_wishlist)))
         onView(withId(R.id.addToWishlist)).perform(click())
         onView(isRoot()).perform(pressBack())
-
+        Thread.sleep(1000)
```
cronokirby, line 55: Why is this necessary?

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -46,8 +52,8 @@ class DetailedItemFragment : Fragment() {
         savedInstanceState: Bundle?
     ): View {
         binding = DataBindingUtil.inflate(inflater, R.layout.fragment_detailed_item, container, false)
-
-        binding.item = args.item
+        getItem()
+        //binding.item = args.item
```
cronokirby, line 56:
```suggestion
```
Let's just remove this comment

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -29,14 +31,18 @@ class DetailedItemFragment : Fragment() {
     private val args: DetailedItemFragmentArgs by navArgs()
     @Inject
     lateinit var currentUserProvider: CurrentUserProvider
+    @Inject
+    lateinit var itemRepository: ItemRepository
     private val viewModel: UserProfileViewModel by viewModels()
     private val itemViewModel: ItemsViewModel by viewModels()
     private lateinit var binding: FragmentDetailedItemBinding
+    private var item: Item? = null
```
cronokirby, line 39: This works, but if the fragment gets reset, because you turn the device, then this gets destroyed. Using part of the ViewModel is really what we should be doing instead.

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -73,6 +79,62 @@ class DetailedItemFragment : Fragment() {
...
+        super.onPrepareOptionsMenu(menu)
+        val sell = menu.findItem(R.id.menuSell)
+        val resell = menu.findItem(R.id.menuResell)
+        if (!args.item.userId.equals(currentUserProvider.getCurrentUserId())) {
+            menu.findItem(R.id.menuEdit).isVisible = false
+            sell.isVisible = false
+            resell.isVisible = false
+        } else {
+            sell.isVisible = !(item?.sold ?: args.item.sold)
+            resell.isVisible = item?.sold ?: args.item.sold
```
cronokirby, line 97:
```suggestion
            resell.isVisible = item?.sold ?: args.item.sold
            sell.isVisible = !resell.isVisible
```

How about this, to avoid repeating the logic?

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -73,6 +79,62 @@ class DetailedItemFragment : Fragment() {
...
+            R.id.menuSell -> {
+                updateSold()
+                true
+            }
+            R.id.menuResell -> {
+                updateSold()
+                true
+            }
+            else -> super.onOptionsItemSelected(item)
+        }
```
cronokirby, line 118: Good use of when btw! Note that you can put multiple conditions, to trigger the block if any of them is true:
```suggestion
        return when (item.itemId) {
            R.id.menuEdit -> {
                view?.findNavController()?.navigate(
                    DetailedItemFragmentDirections.actionDetailedItemFragmentToNewEditFragment(args.item)
                )
                true
            }
            R.id.menuSell, R.id.menuResell -> {
                updateSold()
                true
            }
            else -> super.onOptionsItemSelected(item)
        }
```

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -73,6 +79,62 @@ class DetailedItemFragment : Fragment() {
...
+
+    private fun getItem() {
+        lifecycleScope.launch(Dispatchers.IO) {
+            item = itemRepository.get(args.item.id!!)
+            lifecycleScope.launch(Dispatchers.Main) {
+                binding.item = item
+                activity?.invalidateOptionsMenu()
+            }
+        }
+    }
```
cronokirby, line 136: Is this ever used outside of a coroutine scope (like it is in `updateSold`). If so, consider just making this `private suspend fun getItem()`, and then just calling it directly inside of `updateSold`, to avoid needlessly launching an additional scope

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -81,7 +143,7 @@ class DetailedItemFragment : Fragment() {
             binding.addToWishlist.setOnClickListener { updateWishlist(binding) }
             viewModel.wishlistContains(args.item)
         }else{
```
cronokirby, line 145: This file looks unformatted :P

app/src/main/java/com/example/sharingang/items/ItemsAdapter.kt
```diff
@@ -53,11 +50,11 @@ class ItemsDiffCallback : DiffUtil.ItemCallback<Item>() {
 
 // Replace the description with its ID once the database is set
 class ItemListener(
-    val onEditListener: (item: Item) -> Unit,
+    //val onEditListener: (item: Item) -> Unit,
```
cronokirby, line 53: We should remove all of these comments, no?

app/src/main/java/com/example/sharingang/items/ItemsAdapter.kt
```diff
@@ -28,7 +25,7 @@ class ItemsAdapter(private val clickListener: ItemListener, private val userId:
             binding.itemListViewTitle.text = item.title
             binding.item = item
             binding.clickListener = clickListener
-            binding.isVisible = item.userId?.equals(userId) ?: false
+            //binding.isVisible = item.userId?.equals(userId) ?: false
```
cronokirby, line 28: Remove this comment?

cronokirby: (well, the commented out code)

app/src/main/res/drawable/ic_baseline_money_off_white_24.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="24dp" android:tint="#FFFFFF"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="@android:color/white" android:pathData="M12.5,6.9c1.78,0 2.44,0.85 2.5,2.1h2.21c-0.07,-1.72 -1.12,-3.3 -3.21,-3.81V3h-3v2.16c-0.53,0.12 -1.03,0.3 -1.48,0.54l1.47,1.47c0.41,-0.17 0.91,-0.27 1.51,-0.27zM5.33,4.06L4.06,5.33 7.5,8.77c0,2.08 1.56,3.21 3.91,3.91l3.51,3.51c-0.34,0.48 -1.05,0.91 -2.42,0.91 -2.06,0 -2.87,-0.92 -2.98,-2.1h-2.2c0.12,2.19 1.76,3.42 3.68,3.83V21h3v-2.15c0.96,-0.18 1.82,-0.55 2.45,-1.12l2.22,2.22 1.27,-1.27L5.33,4.06z"/>
```
cronokirby, line 4: I think it should be -1.13l2.23 instead of -1.12l2.22
```
























```
Just kidding 8^)

cronokirby (approved):
Just a few minor improvements to be made, logic seems pretty good otherwise, besides the `var item: Item?` which looks fishy to me

cronokirby (changes requested):
Uh, see my previous review, I uh, selected the wrong option last time :M

app/src/androidTest/java/com/example/sharingang/WishlistViewTest.kt
bafavre: For some reason when removing the buttons, when the list was updating, it would find 2 items with the same id/text and the test would crash. I'm just waiting for the list to update before continuing.

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: Good catch!

app/src/androidTest/java/com/example/sharingang/WishlistViewTest.kt
cronokirby: Interesting; gotta do what you gotta do

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: It was not possible to use the viewmodel here, I was getting errors in the coroutine call mentioning "addObserver" for some reason.

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: Sure good idea!

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: Oh nice, didn't know about that!

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: Yeah that's a good point, I'll see if it works that way because it's indeed better.

app/src/main/java/com/example/sharingang/items/ItemsAdapter.kt
bafavre: My bad, I forgot to remove it apparently.

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: It is called directly in onCreateView, so outside a coroutine scope.

cronokirby: F

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: But when the screen is rotated, on create is called again and it gets the item once again.

cronokirby: Ah ok, that works out then

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -73,32 +82,84 @@ class DetailedItemFragment : Fragment() {
...
+
+    private fun updateSold() {
+        lifecycleScope.launch(Dispatchers.IO) {
+            itemViewModel.sellItem(item)
+            getItem()
+        }
+    }
+
+    private fun getItem() {
+        Log.d("SELL_STATUS", "nice")
```
betrisey, line 128: Maybe loadItem would be a better name, getItem looks like a getter that should return the item.
Also there is a nice debug log 😆

betrisey (approved):
Nice feature and very good implementation :)

I haven't found any other issue

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
bafavre: Whoops, my bad 😆

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -73,32 +81,83 @@ class DetailedItemFragment : Fragment() {
...
+                view?.findNavController()?.navigate(
+                    DetailedItemFragmentDirections.actionDetailedItemFragmentToNewEditFragment(args.item)
+                )
+                true
+            }
+            R.id.menuSell, R.id.menuResell -> {
+                updateSold()
+                true
+            }
+            else -> super.onOptionsItemSelected(item)
```
cronokirby, line 115: noice

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -73,32 +81,83 @@ class DetailedItemFragment : Fragment() {
...
+        super.onPrepareOptionsMenu(menu)
+        val sell = menu.findItem(R.id.menuSell)
+        val resell = menu.findItem(R.id.menuResell)
+        if (!args.item.userId.equals(currentUserProvider.getCurrentUserId())) {
+            menu.findItem(R.id.menuEdit).isVisible = false
+            sell.isVisible = false
+            resell.isVisible = false
+        } else {
+            resell.isVisible = item?.sold ?: args.item.sold
+            sell.isVisible = !resell.isVisible
```
cronokirby, line 99: noice

cronokirby (approved):
noice

PR #175, opened 2021-04-28
Loading spinner and save status for edit/add item

betrisey:
I added a progress bar on top and it displays a snakebar at the end.
Using the Firebase emulator locally, the saving happens a bit too fast to see but it will be more useful when we'll save the images on the remote server.

At the same time, I updated the ItemRepository to merge the `add` and `update` into a single `set` function.

https://user-images.githubusercontent.com/1939395/116434042-59a58c80-a84a-11eb-8207-33a73f76d6c3.mp4


bafavre (approved):
Very good addition, well done! :)

lumapools (approved):
Awesome!

PR #176, opened 2021-04-29
Bug: crashing when opening detailed item

cronokirby:
This is because accessing the arguments can modify a hashmap, so you
must always access arguments from the main / ui thread. This makes sure
this happens in the `loadItem` method.


bafavre (approved):
Nice catch, good job! :)

v1ruso (approved):
Nice, thanks!

PR #177, opened 2021-04-29
View images in grid format

### LINKED ISSUE #155: Browsing item images
As a Buyer, I want to browse different items by seeing their images at a first glance, to judge them better

### LINKED ISSUE #171: Allow browsing item images on list
#156

(Blocked on seeing other images)

cronokirby:
Fixes #171 
Fixes #155 
image
This setups up items to displayed in a nice grid. For now, only placeholder images are used.
Whoever fixes our existing image infrastructure can easily change this code to display the actual image associated with an item, instead of a placeholder. Requesting for permissions is difficult, because when we inflate our view of an item, I don't think we're in any context, so using our permission infra for images is tricky.

### DIFF
--- a/app/src/main/java/com/example/sharingang/ItemsListFragment.kt
+++ b/app/src/main/java/com/example/sharingang/ItemsListFragment.kt
@@ -8,11 +8,14 @@ import androidx.databinding.DataBindingUtil
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.activityViewModels
 import androidx.navigation.fragment.findNavController
+import androidx.recyclerview.widget.GridLayoutManager
 import com.example.sharingang.databinding.FragmentItemsListBinding
 import com.example.sharingang.items.ItemsViewModel
 import com.example.sharingang.users.CurrentUserProvider
 import dagger.hilt.android.AndroidEntryPoint
 import javax.inject.Inject
+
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
```diff
@@ -67,20 +71,22 @@ class DetailedItemFragment : Fragment() {
...
-                    )
-                )
-            }
-        })
+        viewModel.user.observe(viewLifecycleOwner, this::onUserChange)
         return binding.root
     }
 
+    private fun onUserChange(user: User?) {
+        binding.username = "Posted by ${user?.name}"
```
v1ruso, line 79: Is there any way you can use a resource string here? In case we decide to change the language for example, we wouldn't need to go through the code.

v1ruso (approved):
I'm surprised by how little code was needed to implement this feature, well done :)

app/src/main/java/com/example/sharingang/DetailedItemFragment.kt
cronokirby: I didn't add this code, I just formatted the file

betrisey (approved):
Good job!

PR #180, opened 2021-04-29
Return to Detail after Editing and Fix Recursive Navigation

### LINKED ISSUE #179: Improve navigation

bafavre:
Fixes #179 

Now after editing we go back to the detailed view that is updated with the new values.
Also we don't have recursive navigation between the user profile and the detailed view anymore.


betrisey (approved):
Perfect!

Superdoude1999 (approved):
I hate you for the extra merge but it looks good so gj

PR #191, opened 2021-05-01
174 order items

v1ruso:
I implemented a way to order the items on the main page.

For now, we can order them by: Date, Price, Name and Category. We can also order them in descending or ascending order.
Screenshot 2021-05-01 at 12 12 36
Screenshot 2021-05-01 at 12 12 06
Screenshot 2021-05-01 at 12 11 39


lumapools (changes requested):
Very nice, looks good!
I would just have some little "aesthetic" request: I think either the button should have the text "Order" instead of "Order By" as it comes after the two fields, or we can keep "Order By" but put it before the attribute and the direction choices. So that we get either
`Order by` attribute direction
or
attribute direction `Order`
They seem more intuitive than
attribute direction `Order By`

lumapools (approved):
Perfect!

app/src/androidTest/java/com/example/sharingang/OrderItemsTest.kt
```diff
@@ -0,0 +1,111 @@
...
+            typeText(first),
+            closeSoftKeyboard()
+        )
+        onView(withId(R.id.category_spinner)).perform(click())
+        onView(withText(book)).perform(click())
+        onView(withId(R.id.itemPrice)).perform(
+            typeText(firstPrice.toString()),
+            closeSoftKeyboard()
+        )
+        onView(withId(R.id.createItemButton)).perform(click())
```
cronokirby, line 54: You can turn this into a helper function

app/src/androidTest/java/com/example/sharingang/OrderItemsTest.kt
```diff
@@ -0,0 +1,111 @@
...
+        onView(withId(R.id.itemPrice)).perform(
+            typeText(thirdPrice.toString()),
+            closeSoftKeyboard()
+        )
+        onView(withId(R.id.createItemButton)).perform(click())
+
+        onView(withId(R.id.order_category_spinner)).perform(click())
+        onView(withText(orderByName)).perform(click())
+        onView(withId(R.id.start_ordering)).perform(click())
+        Thread.sleep(500)
```
cronokirby, line 87: I really don't see why this would be necessary?

Are you sure that this is necessary to make tests pass, or is this just a 'hail mary'?

app/src/main/java/com/example/sharingang/utils/UIViewModel.kt
```diff
@@ -0,0 +1,26 @@
+package com.example.sharingang.utils
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+class UIViewModel : ViewModel() {
```
cronokirby, line 7: This is a really vaguely named class: "UIViewModel" could apply to any of our ViewModels: all of them are about the UI.

(It would also be a good idea to document this class; let's get into the habit of doing that)

app/src/main/java/com/example/sharingang/utils/UIViewModel.kt
```diff
@@ -0,0 +1,26 @@
...
+    val orderByPosition: LiveData<Int>
+        get() = _orderByPosition
+
+    fun setAscendingDescendingPosition(position: Int) {
+        _ascendingDescendingPosition.postValue(position)
+    }
+
+    fun setOrderByPosition(position: Int) {
+        _orderByPosition.postValue(position)
+    }
```
cronokirby, line 25: Are these called from a coroutine?

You should just do `_orderByPosition.value = position` to normally set the LiveData. `postValue` is only necessary of something is being called from a coroutine.

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
```diff
@@ -127,6 +135,29 @@ class ItemsViewModel @Inject constructor(
...
+                    ORDERING.DATE -> it.createdAt
+                    ORDERING.PRICE -> it.price
+                    ORDERING.NAME -> it.title
+                    ORDERING.CATEGORY -> it.category
+                }
+            })
+            if (!isAscending) {
+                results = results.asReversed()
+            }
+            _orderedItems.postValue(results)
```
cronokirby, line 157: Is this called from a coroutine? `postValue` isn't necessary if not

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -80,14 +80,17 @@ class MainActivity : AppCompatActivity(R.layout.activity_main) {
...
+            if (allItems.isNotEmpty()) {
+                val random = allItems.random()
+                lifecycleScope.launch(Dispatchers.Main) {
+                    val drawerLayout = findViewById<DrawerLayout>(R.id.drawer_layout)
+                    drawerLayout.close()
+                    getNavController().navigate(
+                        ItemsListFragmentDirections
+                            .actionItemsListFragmentToDetailedItemFragment(random)
+                    )
+                }
```
cronokirby, line 93:
```suggestion
            val allItems = itemRepository.getAll()
            if (allItems.isEmpty()) {
              return
            }
            val random = allItems.random()
            lifecycleScope.launch(Dispatchers.Main) {
              val drawerLayout = findViewById<DrawerLayout>(R.id.drawer_layout)
              drawerLayout.close()
              getNavController().navigate(
                ItemsListFragmentDirections
                  .actionItemsListFragmentToDetailedItemFragment(random)
              )
            }
```

Let's just use an early return here instead

app/src/main/java/com/example/sharingang/ItemsListFragment.kt
```diff
@@ -20,32 +21,48 @@ class ItemsListFragment : Fragment() {
...
 
         binding.swiperefresh.setOnRefreshListener { viewModel.refresh() }
         viewModel.refreshing.observe(viewLifecycleOwner, {
             if (!it) {
                 binding.swiperefresh.isRefreshing = false
+                orderItems() // in the case an item is added, need to add redo the ordering
             }
         })
+        binding.orderCategorySpinner.setSelection(uiViewModel.orderByPosition.value!!)
+        binding.orderAscendingDescending.setSelection(uiViewModel.ascendingDescendingPosition.value!!)
```
cronokirby, line 52: Are the `!!` necessary here? Both of the livedata don't contain null values, so this seems odd

cronokirby (changes requested):
Just a handful of things to look at, otherwise GG

app/src/androidTest/java/com/example/sharingang/OrderItemsTest.kt
v1ruso: Very good point

app/src/androidTest/java/com/example/sharingang/OrderItemsTest.kt
v1ruso: It was necessary, the update of the UI is not instantaneous

app/src/main/java/com/example/sharingang/utils/UIViewModel.kt
v1ruso: True, well I thought we could use this to keep specific UI elements in memory, like the spinners selected positions. Since it could also be used for other classes, this is the more general name I could come up with. Anything you would recommend instead?

cronokirby: OrderingViewModel?

app/src/main/java/com/example/sharingang/utils/UIViewModel.kt
v1ruso: Thanks

app/src/main/java/com/example/sharingang/items/ItemsViewModel.kt
v1ruso: I copied the structure from searchItems

cronokirby: Yeah, this is never called from inside a coroutine, so you can just do `_orderedItems.value = results`

v1ruso: And at this point we're in `viewModelScope.launch(Dispatchers.IO) { ... }` which would make it inside a coroutine

cronokirby: Oh, right

app/src/main/java/com/example/sharingang/MainActivity.kt
v1ruso: Not allowed apparently

app/src/main/java/com/example/sharingang/ItemsListFragment.kt
v1ruso: Yeah I was confused too, Android Studio keeps telling me it needs that

app/src/main/java/com/example/sharingang/utils/UIViewModel.kt
v1ruso: Perfect

cronokirby (approved):
cool beans

PR #197, opened 2021-05-05
Add shaking to randomize

### LINKED ISSUE #188: Shake to randomize
As a Buyer, I want to shake my phone to randomize the random item page, because I like having fun

cronokirby:
Fixes #188


app/src/main/java/com/example/sharingang/shake/ShakeDetector.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun detect(x: Float, y: Float, z: Float, now: Long): Boolean {
+        val gForce = sqrt(x * x + y * y + z * z) / SensorManager.GRAVITY_EARTH
+        // Ignore shakes without enough force
+        if (gForce < SHAKE_THRESHOLD_GRAVITY) {
+            return false
+        }
+        // Ignore shakes that happen too close to each other
+        if (shakeTimestamp + SHAKE_WAIT > now) {
+            return false
+        }
```
lumapools, line 45: I think you could merge these two `if`s

app/src/test/java/com/example/sharingang/shake/ShakeDetectorTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import org.junit.Test
+
+class ShakeDetectorTest {
+    @Test
+    fun noShakeIsDetectedForSmallMovements() {
+        val detector = ShakeDetector()
+        assert(!detector.detect(0F, 0F, 0F, 0))
+        assert(!detector.detect(0F, 0F, 0F, 10000))
+    }
```
lumapools, line 11: Does this test case cover the case when there is no movement at all? I think you could maybe add a test case for when there is movement but not enough to trigger the shake condition.

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -78,17 +106,23 @@ class MainActivity : AppCompatActivity(R.layout.activity_main) {
...
-                val random = allItems.random()
+
+                val random = itemRepository.getAll().random()
                 lifecycleScope.launch(Dispatchers.Main) {
                     val drawerLayout = findViewById<DrawerLayout>(R.id.drawer_layout)
                     drawerLayout.close()
                     getNavController().navigate(
                         ItemsListFragmentDirections
-                            .actionItemsListFragmentToDetailedItemFragment(random)
+                            .actionGlobalDetailedItemFragment(random)
```
lumapools, line 125: I am not sure I understand this change. What is the difference between `action...To...` and `actionGlobal...`?

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -78,17 +106,23 @@ class MainActivity : AppCompatActivity(R.layout.activity_main) {
...
+        selectRandomItem()
+    }
+
+    private fun selectRandomItem() {
         lifecycleScope.launch(Dispatchers.IO) {
             val allItems = itemRepository.getAll()
             if (allItems.isNotEmpty()) {
-                val random = allItems.random()
+
+                val random = itemRepository.getAll().random()
```
lumapools, line 119: Why do we need to `getAll()` again instead of keeping `allItems.random()`?

lumapools (changes requested):
Awesome, well done! 
There are just a few things that I did not fully understand but this definitely is a fun feature! ^^

Edit: Oops I just noticed this is a Draft, I don't know if it's fine to request changes in a WIP.

app/src/main/java/com/example/sharingang/MainActivity.kt
v1ruso: Not sure I understand why either

app/src/main/java/com/example/sharingang/MainActivity.kt
```diff
@@ -24,6 +28,10 @@ class MainActivity : AppCompatActivity(R.layout.activity_main) {
     @Inject
     lateinit var itemRepository: ItemRepository
 
+    lateinit var shakeListener: ShakeListener
+    lateinit var sensorManager: SensorManager
+    lateinit var accelerometer: Sensor
```
v1ruso, line 33: Do you need this to be public?

app/src/test/java/com/example/sharingang/shake/ShakeDetectorTest.kt
cronokirby: Yeah, not a bad idea for small movement

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: The former requires us to be on the items list fragment. This crashes if you're not there, e.g. shaking from somewhere else. Using a global action makes us navigate regardless of where we happen to be.

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: woops

app/src/main/java/com/example/sharingang/shake/ShakeDetector.kt
cronokirby: Sure, but I feel like it's clearer if the logic is separate

app/src/main/java/com/example/sharingang/MainActivity.kt
cronokirby: Mistake when merging :P

app/src/main/java/com/example/sharingang/MainActivity.kt
lumapools: Okay, thanks!

lumapools (approved):
Nice!

v1ruso (approved):
Well done:)

PR #206, opened 2021-05-07
Subscribe to category

### LINKED ISSUE #198: Add possibility to subscribe to a category

bafavre:
Fixes #198 

Now we can subscribe to categories to later receive push notifications when a new item is added.


app/src/main/java/com/example/sharingang/utils/NotificationUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+    contentPendingIntent: PendingIntent
+): NotificationCompat.Builder {
+    return NotificationCompat.Builder(
+        applicationContext,
+        applicationContext.getString(R.string.new_item_notification_channel_id)
+    ).setSmallIcon(R.drawable.ic_launcher_foreground)
+        .setContentTitle(
+            applicationContext.getString(
+                R.string.new_item_notification_message,
+                "test"
```
betrisey, line 36: I guess that it will be updated when you will send real notifications from the server

betrisey (approved):
Very nice, as always :)

app/src/main/java/com/example/sharingang/utils/NotificationUtils.kt
bafavre: Yes indeed

app/src/main/java/com/example/sharingang/UserProfileViewModel.kt
```diff
@@ -90,6 +94,46 @@ class UserProfileViewModel @Inject constructor(
         }
     }
 
+    /**
+     * Check whether the connected user is subscribed
+     * to this category's notifications
+     * @param category the String of the category
+     * @return whether the set contains this category already
+     */
```
Superdoude1999, line 102: wow there is even documentation... much impressive

Superdoude1999 (approved):
nice job

PR #214, opened 2021-05-10
Subscription View

### LINKED ISSUE #201: Add a view where I can see what my notification subscriptions are

bafavre:
Fixes #201

We now have a way to see our subscriptions, even if it's not very pretty.


v1ruso (commented):
I haven't looked at the code yet, but would it be possible to unsubscribe from a category directly from this view? Would it be hard to implement?

lumapools (approved):
Looks good, I just have one question about the first item being empty.

PR #230, opened 2021-05-14
Move Cloud functions to Zurich datacenter

betrisey:
Grüezi!
Should be faster because the other Firebase services are hosted in Zurich (europe-west6) 
Also added a check to only load serviceAccountKey.json when running locally (we don't need it on Google Cloud)


Superdoude1999 (approved):
Wir sind nach Zürich gezogen

lumapools (approved):

lumapools:
> Wir sind nach Zürich gezogen

You used Google Translate for this didn't ya

PR #239, opened 2021-05-18
Fix layout, last items row was cropped

betrisey:
The last row is no longer cropped :)


v1ruso (approved):
Nice catch:)

lumapools (approved):
Thanks for the fix!

PR #244, opened 2021-05-18
Improve Profile Picture change

lumapools:
Picture is changed without the need of an Apply button. With `onResume()`, we can now just check there if the image has changed, and, if yes, update the user's profile picture. Otherwise do nothing. There is no more need for an Apply button.


v1ruso (approved):
Great! Much better without the apply button:)

betrisey (approved):
Great :)

PR #249, opened 2021-05-18
Unread messages

lumapools:
The users can now see how many unread messages they have from another particular user. If they open a conversation or are already in a conversation, their unread messages will be 0 (hence the red circle with the number of unread messages won't be shown).

Webp net-resizeimage (9)

The Chats UI is updated whenever they receive a new message (no need to refresh), but if it is a new partner the UI is not refreshed automatically.


lumapools:
Thought it would be fine, but CI doesn't use Firestore. Great...

PR #250, opened 2021-05-19
Add an activity for seeing locating information

cronokirby:
image

This activity contains all of the information needed for doing AR stuff. The idea is to later display a 3d arrow or something that shows you how to walk towards an item, or something like that.

(Hopefully the coverage is good enough)

### DIFF
--- a/app/src/main/java/com/example/sharingang/models/Location.kt
+++ b/app/src/main/java/com/example/sharingang/models/Location.kt
@@ -50,4 +50,21 @@ data class Location(val lat: Double, val lon: Double) {
         val c = 2 * atan2(sqrt(a), sqrt(1 - a))
         return EARTH_RADIUS * c
     }
+
+    /**
+     * This calculates the heading required to orient yourself towards another location.
+     *
+     * @param that the location to orient yourself to, from this location
+     * @return the required heading
+     */
+    fun requiredHeading(that: Location): Heading {
+        val phi1 = degreesToRadians(this.lat)
+        val phi2 = degreesToRadians(that.lat)
+        val deltaLambda = degreesToRadians(that.lon - this.lon)
+
+        val y = sin(deltaLambda) * cos(phi2)
+        val x = cos(phi1) * sin(phi2) - sin(phi1) * cos(phi2) * cos(deltaLambda)
+        val theta = atan2(y, x)
+        return Heading(theta)
+    }
 }
--- a/app/src/main/java/com/example/sharingang/models/Location.kt
+++ b/app/src/main/java/com/example/sharingang/models/Location.kt
@@ -1,0 +1,12 @@
+package com.example.sharingang.models
+
+/**
+ * Location represents a single point on the Earth.
+ *
+ * We use a latitude, and a longitude, both doubles. Now, doubles can be somewhat misleading,
+ * because the precision of the sensor data we receive is actually less accurate.
+ * On the other hand, it can't actually hurt to use more accurate numbers when doing
+ * calculations. We also don't perform many expensive calculations, so the use of Double
+ * isn't very onerous either.
+ */
+data class Location(val lat: Double, val lon: Double)
--- a/app/src/main/java/com/example/sharingang/models/Location.kt
+++ b/app/src/main/java/com/example/sharingang/models/Location.kt
@@ -1,5 +1,16 @@
 package com.example.sharingang.models
 
+import com.example.sharingang.utils.degreesToRadians
+import kotlin.math.atan2
+import kotlin.math.cos
+import kotlin.math.sin
+import kotlin.math.sqrt
+
+/**
+ * The radius of the earth, in meters
+ */
+private const val EARTH_RADIUS: Double = 6371e3
+
 /**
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/example/sharingang/models/Location.kt
```diff
@@ -0,0 +1,70 @@
...
+            sinDeltaPhi * sinDeltaPhi + cos(phi1) * cos(phi2) * sinDeltaLambda * sinDeltaLambda
+        val c = 2 * atan2(sqrt(a), sqrt(1 - a))
+        return EARTH_RADIUS * c
+    }
+
+    /**
+     * This calculates the heading required to orient yourself towards another location.
+     *
+     * @param that the location to orient yourself to, from this location
+     * @return the required heading
```
betrisey, line 58: Is it relative to the angle relative to the north?

app/src/main/res/layout/fragment_detailed_item.xml
```diff
@@ -225,9 +225,18 @@
...
                 app:srcCompat="?attr/actionModeShareDrawable" />
+
+            <com.google.android.material.floatingactionbutton.FloatingActionButton
+                android:id="@+id/locateButton"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_gravity="end|bottom"
+                android:layout_marginTop="16dp"
+                android:contentDescription="@string/locate_item"
+                app:srcCompat="@drawable/baseline_search_black_24dp" />
```
betrisey, line 239: We could use the same style for both buttons. Also it would be better if the icon was white too.
```suggestion
                android:backgroundTint="@color/themeColor"
                android:layout_marginTop="16dp"
                android:contentDescription="@string/share"
                app:srcCompat="?attr/actionModeShareDrawable" />

            <Space
                android:layout_width="10dp"
                android:layout_height="wrap_content" />

            <com.google.android.material.floatingactionbutton.FloatingActionButton
                android:id="@+id/locateButton"
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                android:layout_gravity="end|bottom"
                android:backgroundTint="@color/themeColor"
                android:layout_marginTop="16dp"
                android:contentDescription="@string/locate_item"
                app:srcCompat="@drawable/baseline_search_black_24dp" />
```

betrisey (changes requested):
Very good implementation :)
I've tested on my phone too and it works well.

app/src/main/java/com/example/sharingang/models/Location.kt
cronokirby: I believe so: https://www.wikiwand.com/en/Azimuth

betrisey: Okay thanks

app/src/main/res/layout/fragment_detailed_item.xml
```diff
@@ -225,9 +225,24 @@
...
+
+            <com.google.android.material.floatingactionbutton.FloatingActionButton
+                android:id="@+id/locateButton"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_gravity="end|bottom"
+                android:backgroundTint="@color/themeColor"
+                android:layout_marginTop="16dp"
+                android:contentDescription="@string/locate_item"
+                app:srcCompat="@drawable/baseline_search_black_24dp" />
```
betrisey, line 245: Could you just make the icon white?

betrisey (approved):

app/src/main/res/layout/fragment_detailed_item.xml
cronokirby: image
Replaced with this icon

betrisey: Perfect

app/src/main/java/com/example/sharingang/models/Location.kt
```diff
@@ -0,0 +1,70 @@
...
+     * This is "as the crow flies", in the sense that the topography of the earth isn't
+     * taken into account, and the Earth is treated as a sphere.
+     *
+     * Unfortunately, in practice not only is the Earth not a sphere, but an elongated
+     * ellipsoid, because of its rotation, but there are also troublesome things like
+     * mountains, buildings, hills, etc.
+     *
+     * This distance is still a reasonable approximation.
+     *
+     * @param that the location to calculate the distance to, compared to this location
```
bafavre, line 36: The params would be "lat" and "long" instead of "that" I'd say.

bafavre (approved):
Good job, just a small typo I guess when writing documentation, but well done! :)

app/src/main/java/com/example/sharingang/models/Location.kt
cronokirby: Well, there's only a single parameter `that` of type location, so this seems to be how javadoc should work, I think?

bafavre: Oh yeah my bad, I mixed the lat and long from the data class and thought it was in the function, whoops

PR #251, opened 2021-05-19
Add security rules

v1ruso:
Finally doing some backend stuff.

I added security rules for Firestore: before anyone could do anything on the database. It's now fixed.

### DIFF
--- a/app/src/main/java/com/example/sharingang/database/room/ListConverter.kt
+++ b/app/src/main/java/com/example/sharingang/database/room/ListConverter.kt
@@ -15,7 +15,7 @@ object ListConverter {
      */
     @TypeConverter
     fun toList(stringList: String?): List<String?> {
-        if (stringList == null) {
+        if (stringList.isNullOrEmpty()) {
             return ArrayList()
         }
         return stringList.split(',')
--- a/app/src/main/java/com/example/sharingang/ui/fragments/SubscriptionFragment.kt
+++ b/app/src/main/java/com/example/sharingang/ui/fragments/SubscriptionFragment.kt
@@ -35,7 +35,7 @@ class SubscriptionFragment : Fragment() {
         userViewmodel.setUser(userId)
         userViewmodel.user.observe(viewLifecycleOwner) { user ->
             // More than one since the first item is always empty
-            if (user != null && user.subscriptions.size > 1) {
+            if (user != null) {
                 val message = formatSubscriptions(user.subscriptions)
                 binding.subscriptions.text = message
             }
@@ -46,7 +46,7 @@ class SubscriptionFragment : Fragment() {
 
     private fun formatSubscriptions(subscriptions: List<String>): String {
         var message = context?.getString(R.string.subscribed_to)
-        message = message.plus(TextUtils.join(", ", subscriptions.drop(1))).plus(".")
+        message = message.plus(TextUtils.join(", ", subscriptions)).plus(".")
         return message
     }
 }
--- a/app/src/main/java/com/example/sharingang/ui/fragments/SubscriptionFragment.kt
+++ b/app/src/main/java/com/example/sharingang/ui/fragments/SubscriptionFragment.kt
@@ -34,8 +34,7 @@ class SubscriptionFragment : Fragment() {
         val userId = currentUserProvider.getCurrentUserId()
         userViewmodel.setUser(userId)
         userViewmodel.user.observe(viewLifecycleOwner) { user ->
-            // More than one since the first item is always empty
-            if (user != null) {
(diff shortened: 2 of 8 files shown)


firestore.rules
```diff
@@ -1,8 +1,25 @@
 rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
-    match /{document=**} {
-      allow read, write: if true;
-    }
+        match /items/{itemId} {
+          allow create: if request.auth != null;
+          allow update, delete: if request.auth.uid == request.resource.data.userId;
```
betrisey, line 6: We should also check that the userId on the existing item (`resource.data`) matches
```suggestion
          allow update, delete: if request.auth != null && request.auth.uid == request.resource.data.userId && request.auth.uid == resource.data.userId;
```

firestore.rules
```diff
@@ -1,8 +1,25 @@
 rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
-    match /{document=**} {
-      allow read, write: if true;
-    }
+        match /items/{itemId} {
+          allow create: if request.auth != null;
```
betrisey, line 5: Maybe we could use the save rule as the update, delete. This was are sure that the userId is correct when creating an item

firestore.rules
```diff
@@ -1,8 +1,25 @@
...
+          allow update, delete: if request.auth.uid == request.resource.data.userId;
+          allow read: if true;
+        }
+        match /users/{userId} {
+          allow create, read: if true;
+          allow update: if request.auth != null && request.auth.uid == userId;
```
betrisey, line 11:
```suggestion
          allow read: if true;
          allow create, update: if request.auth != null && request.auth.uid == userId;
```

betrisey (changes requested):
It looks good :)
Could you add rules for the chat, maybe ask @lumapools for the exact structure?

firestore.rules
v1ruso: Mh this can't work, we need to be logged out to create a user

betrisey: When the user logs in with Google, the `auth` value is set and after that we add it to Firestore so it may work. Could you try it please?
https://github.com/Sharingang/Sharingang-Android/blob/5dcab0936429b0b248da78fdb5233a5532a22de5/app/src/main/java/com/example/sharingang/auth/AuthHelper.kt#L38-L42

firestore.rules
v1ruso: Tried, doesn't work...

firestore.rules
v1ruso: You're right, it works

betrisey (approved):
Very good, the rules look correct
When it's merged, you can update them on the Firebase console

app/src/main/java/com/example/sharingang/database/room/ListConverter.kt
```diff
@@ -15,7 +15,7 @@ object ListConverter {
      */
     @TypeConverter
     fun toList(stringList: String?): List<String?> {
-        if (stringList == null) {
+        if (stringList.isNullOrEmpty()) {
```
bafavre, line 18: Nice catch!

app/src/main/java/com/example/sharingang/ui/fragments/SubscriptionFragment.kt
```diff
@@ -34,8 +34,7 @@ class SubscriptionFragment : Fragment() {
         val userId = currentUserProvider.getCurrentUserId()
         userViewmodel.setUser(userId)
         userViewmodel.user.observe(viewLifecycleOwner) { user ->
-            // More than one since the first item is always empty
-            if (user != null && user.subscriptions.size > 1) {
+            if (user != null && user.subscriptions.isNotEmpty()) {
```
bafavre, line 37: Makes more sense with the correct converter :)

app/src/main/java/com/example/sharingang/viewmodels/UserProfileViewModel.kt
```diff
@@ -145,7 +145,7 @@ class UserProfileViewModel @Inject constructor(
     fun loginResubscribe(userId: String) {
         viewModelScope.launch(Dispatchers.IO) {
             val user = userRepository.get(userId)
-            user?.subscriptions?.forEach { if (it.isNotEmpty()) subscribeToTopic(it) }
+            user?.subscriptions?.forEach { subscribeToTopic(it) }
```
bafavre, line 148: Makes more sense with the correct converter :)

bafavre (approved):
Good job, well done for the converter, I didn't notice it! :)

PR #252, opened 2021-05-19
Fix picture buttons color (UserProfile)

lumapools:
(Too lazy to rescale image)

The buttons were sometimes blending with the picture due to their bright/dark; now the color is fixed and won't blend anymore.

Screenshot from 2021-05-19 22-24-09


v1ruso (approved):
Maybe could you do like it is in New item? With a lighter button background and smaller icons?

betrisey (approved):

PR #258, opened 2021-05-20
Discounts

### LINKED ISSUE #240: Discounts
As a Seller, I want to be able to advertize a sale on an item, to encourage people to buy it

### LINKED ISSUE #257: Add possibility to discount an item

bafavre:
Fixes #240 
Fixes #257

Users can now set an item as being discounted, so other users might be more interested in it. The discount price field is only shown when the discount switch is checked. There is also a validation that happens to check that the discount price is lower than the normal price.
Also on the detailed view of an item, we see the old price crossed and the new price next to it.


Superdoude1999 (approved):
Looks fine to me

v1ruso (approved):
Looks good to me

PR #268, opened 2021-05-21
Last Item Update

lumapools:
The item view shows how much time ago the item has been created 
(or updated if it has been updated since its creation).  
It is also shown on the Detailed View.

1. 55 seconds (left) and 2 minutes (right) after the upload. (See more info below the pictures) 
Webp net-resizeimage (14) Webp net-resizeimage (15)

2. Detailed View: 
Webp net-resizeimage (16)

The timestamps are seconds, minutes, hours, days, weeks, months and years (shown respectively as 
`s`, `min`, `h`, `d`, `w`, `mon` and `y`). If the item has been posted less than a minute ago, it will show seconds, if the item has been posted more than a minute ago but less than an hour ago, it will show minutes, etc...

Notes:
1. There are some imprecisions in the code concerning the calculations (i.e. a year does not always have 365 days and a month does not always contain 30 days), but as we are using these as estimates for the buyer to have an idea about when the item has last been updated, I think it is fine this way.
2. I cannot make the text or the icon bigger otherwise it will either not show completely, or cover up the item image.

### DIFF
--- a/app/src/main/java/com/example/sharingang/database/cache/CachedItemRepository.kt
+++ b/app/src/main/java/com/example/sharingang/database/cache/CachedItemRepository.kt
@@ -7,7 +7,7 @@ import com.example.sharingang.database.repositories.ItemRepository
 import com.example.sharingang.database.store.ItemStore
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
-import java.util.*
+import java.util.Date
 import javax.inject.Inject
 
 
--- a/app/src/main/java/com/example/sharingang/database/cache/CachedItemRepository.kt
+++ b/app/src/main/java/com/example/sharingang/database/cache/CachedItemRepository.kt
@@ -7,6 +7,7 @@ import com.example.sharingang.database.repositories.ItemRepository
 import com.example.sharingang.database.store.ItemStore
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import java.util.*
 import javax.inject.Inject
 
 
@@ -61,6 +62,14 @@ class CachedItemRepository @Inject constructor(
 
     override suspend fun delete(id: String): Boolean {
         return thenRefresh { store.delete(id) }
+    }
+
+    override suspend fun getLastTimeUpdate(id: String): Date {
+        return thenRefresh { store.getLastTimeUpdate(id) }
+    }
+
+    override suspend fun setLastTimeUpdate(id: String, newValue: Date) {
+        return thenRefresh { store.setLastTimeUpdate(id, newValue) }
     }
 }
 
--- a/app/src/main/java/com/example/sharingang/database/firestore/FirestoreItemStore.kt
+++ b/app/src/main/java/com/example/sharingang/database/firestore/FirestoreItemStore.kt
@@ -2,7 +2,10 @@ package com.example.sharingang.database.firestore
 
 import com.example.sharingang.models.Item
 import com.example.sharingang.database.store.ItemStore
+import com.example.sharingang.utils.constants.DatabaseFields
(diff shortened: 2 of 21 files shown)


v1ruso (commented):
Can you add a fine grained precision? so show minutes (maybe even seconds) if under an hour, hours if under a day, and days otherwise?

v1ruso (approved):
Looks good! Do the timestamps update in real time?

lumapools:
Unfortunately they don't udpate in real time, but I think it is not a big issue, as we don't want to constantly poll the database to check the update time of all items (so we won't see 1s, 2s, 3s, ... changing when we are on the item list for example), plus it already gives an accurate estimate to the user navigating through activities. The only use of this would be if someone left their phone on one particular activity for months (in this case the timestamps wouldn't be updated), but it is an uncommon / weird scenario that is not worth handling. If an activity containing an item element is refreshed (or if we leave it and come back to it) the timestamp on each item is updated of course.

bafavre (commented):
I haven't looked at the code yet, but doesn't it seem a bit too much? I don't think we need to see the time in the list directly, but only on the detailed page, don't you think?
Also having the time difference since the last change also seems too much, couldn't we just show the last time it was updated as a time? For example up to the minute like 2021-05-25 12:25
This seems like a messaging part when it should only give an indication to the user of when it was last updated.
What do you think?

app/src/main/java/com/example/sharingang/database/cache/CachedItemRepository.kt
```diff
@@ -7,6 +7,7 @@ import com.example.sharingang.database.repositories.ItemRepository
 import com.example.sharingang.database.store.ItemStore
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import java.util.*
```
cronokirby, line 10: Maybe just import `java.util.Date`?

app/src/main/java/com/example/sharingang/database/firestore/FirestoreItemStore.kt
```diff
@@ -26,4 +29,14 @@ class FirestoreItemStore @Inject constructor(firestore: FirebaseFirestore) :
             }
         }
     }
+
+    override suspend fun getLastTimeUpdate(id: String): Date {
+        return firestore.collection(DatabaseFields.DBFIELD_ITEMS)
+            .document(id).get().await().getDate(DatabaseFields.DBFIELD_UPDATED_AT)!!
```
cronokirby, line 35: What if the document was never update before?

app/src/main/java/com/example/sharingang/ui/fragments/DetailedItemFragment.kt
```diff
@@ -249,7 +253,7 @@ class DetailedItemFragment : Fragment() {
         val shareIntent = Intent.createChooser(Intent().apply {
             action = Intent.ACTION_SEND
             putExtra(Intent.EXTRA_TEXT, link.toString())
-            putExtra(Intent.EXTRA_TITLE, generateLinkTitle(args.item))
+            putExtra(Intent.EXTRA_TITLE, args.item.title + " - " + getString(R.string.app_name))
```
cronokirby, line 256: Use a formatting string here instead of concatenation, imo

cronokirby (approved):
Looks good to me, just some minor changes you might potentially want to look at.

app/src/main/java/com/example/sharingang/database/firestore/FirestoreItemStore.kt
lumapools: When the item is created, it is considered to be an update, therefore the "updatedAt" field will be non-null.

PR #270, opened 2021-05-22
Message timestamps

lumapools:
Messages have now timestamps (more information below the picture).

Webp net-resizeimage (13)

If a message has been sent today, it will have `Today, hh:mm` as timestamp. Ex: `Today, 12:03` 
(If the hour is less than 10, we just have `h:mm`)
If a message has been sent this year but not today, it will have `monthName. dd` as timestamp. Ex: `Feb. 2` 
Otherwise, it will have `monthName. dd, yyyy` as timestamp. Ex: `Mar. 12, 2020`

Where

`monthName` corresponds to the first 3 letters of the month in question. We don't want too much text under the message, it should just be an indicator, due to which 3 letters are fine in my opinion.

### DIFF
--- a/app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
+++ b/app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
@@ -9,6 +9,7 @@ import androidx.recyclerview.widget.RecyclerView
 import com.example.sharingang.R
 import com.example.sharingang.models.Chat
 import com.example.sharingang.ui.fragments.MessageFragment
+import com.example.sharingang.utils.DateHelper
 
 /**
  * MessageAdapter takes care of adapting a list of messages into a Recycler View.
@@ -56,7 +57,7 @@ class MessageAdapter(
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
         val message: String = chats[position].message
         holder.text.text = message
-        holder.time.text = chats[position].getMessageTime()
+        holder.time.text = DateHelper.formatMessageDate(chats[position].date)
     }
 
     override fun getItemCount(): Int {
--- a/app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
+++ b/app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
@@ -19,7 +19,7 @@ import com.example.sharingang.utils.DateHelper
  * @property currentUserId the current logged in user's id
  */
 class MessageAdapter(
-    private val context: Context, private var chats: List<Chat>,
+    private val context: Context, private var chats: MutableList<Chat>,
     private val currentUserId: String, private val attachedFragment: MessageFragment
 ) :
     RecyclerView.Adapter<MessageAdapter.ViewHolder>() {
@@ -78,7 +78,8 @@ class MessageAdapter(
      * @param newData the incoming data
      */
     fun submitList(newData: List<Chat>) {
-        chats = newData
+        chats.clear()
+        chats.addAll(newData)
         notifyDataSetChanged()
         attachedFragment.scrollToEnd()
     }
--- a/app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
+++ b/app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
@@ -18,7 +18,7 @@ import com.example.sharingang.ui.fragments.MessageFragment
  * @property currentUserId the current logged in user's id
  */
 class MessageAdapter(
-    private val context: Context, private var chats: MutableList<Chat>,
+    private val context: Context, private var chats: List<Chat>,
     private val currentUserId: String, private val attachedFragment: MessageFragment
 ) :
     RecyclerView.Adapter<MessageAdapter.ViewHolder>() {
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
```diff
@@ -75,8 +78,7 @@ class MessageAdapter(
      * @param newData the incoming data
      */
     fun submitList(newData: List<Chat>) {
-        chats.clear()
-        chats.addAll(newData)
+        chats = newData
```
v1ruso, line 81: Can't this be a problem if the parameter `newData` is changed afterwards?

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
```diff
@@ -0,0 +1,43 @@
...
+        2 to "Mar",
+        3 to "Apr",
+        4 to "May",
+        5 to "Jun",
+        6 to "Jul",
+        7 to "Aug",
+        8 to "Sep",
+        9 to "Oct",
+        10 to "Nov",
+        11 to "Dec"
```
v1ruso, line 19: Could you use a string resource instead?

app/src/main/java/com/example/sharingang/ui/fragments/MessageFragment.kt
```diff
@@ -63,10 +63,10 @@ class MessageFragment : Fragment() {
         setupFields()
         setupSendButton()
         setupUI()
-        messagesLiveData.postValue(listChats)
         lifecycleScope.launch(Dispatchers.Main) {
             setupMessageRefresh()
         }
+        messagesLiveData.postValue(listChats)
```
v1ruso, line 69: I don't think you need postValue, as this seems to be executed in the main thread, see [https://stackoverflow.com/questions/51299641/difference-of-setvalue-postvalue-in-mutablelivedata](url).
```suggestion
        messagesLiveData.setValue(listChats)
```

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
```diff
@@ -0,0 +1,43 @@
...
+     * @return the formatted date
+     */
+    fun formatMessageDate(date: Date): String {
+        val currentDate = Date()
+        val today = Triple(currentDate.year, currentDate.month, currentDate.day)
+        val messageTime = Triple(date.year, date.month, date.day)
+        val monthStr = monthsMap[date.month]
+        return (
+            when {
+                today == messageTime -> "Today, ${date.hours}:${
```
v1ruso, line 35: String resource? Would be easier for possible future translation

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
lumapools: I don't know how to do that, as string resources can only be accessed only from a Context.

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
lumapools: If you know a way of accessing the string resource file without a context I would gladly change. However I don't know how to do that.

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
v1ruso: [https://stackoverflow.com/questions/4253328/getstring-outside-of-a-context-or-activity](url)

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
v1ruso: [https://stackoverflow.com/questions/4253328/getstring-outside-of-a-context-or-activity](url)

app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
lumapools: If the newData changes, it means that there was a change in the messages. If there is a change in the messages, it means that there is a listener that will see that change and update the view. Therefore I don't think it is a problem.

v1ruso: Ok great!

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
lumapools: Thanks! The website says 404 not found but I will look it up 👌🏻

app/src/main/java/com/example/sharingang/ui/adapters/MessageAdapter.kt
lumapools: However I needed to revert some changes as there was some issue with the updates so now it's back at how it was before 😅

app/src/main/java/com/example/sharingang/utils/DateHelper.kt
lumapools: I have just looked up the solutions, however, none of them could solve the problem. Apparently, only the system resources are accessible with the suggested method, not the ones bound to the application.

v1ruso: Could we use system resources then?

lumapools: The system resources are not modifiable because they are provided by the SDK, not the app. I don't think there was any other solution than using `DateHelper` as a Class, not an Object, and providing it with the context. So that's what I did

v1ruso (approved):
Great, well done!

PR #275, opened 2021-05-25
Mark discounted items

### LINKED ISSUE #273: Change UI for discounted items.

bafavre:
Fixes #273 
We now have a small indication that an item has been discounted.


v1ruso (approved):
Great work

v1ruso (commented):
Could you add a padding so that the circle is a bit detached from the rest? And maybe make the circle a bit smaller?

Superdoude1999 (approved):
Simple but efficient

PR #279, opened 2021-05-25
Notifications chat

v1ruso:
Now users receive a notification when other people text them!
Screenshot 2021-05-26 at 12 01 24

### DIFF
--- a/app/src/main/java/com/example/sharingang/utils/notification/MyFirebaseMessagingService.kt
+++ b/app/src/main/java/com/example/sharingang/utils/notification/MyFirebaseMessagingService.kt
@@ -2,6 +2,7 @@ package com.example.sharingang.utils.notification
 
 import android.app.NotificationManager
 import androidx.core.content.ContextCompat
+import com.example.sharingang.R
 import com.example.sharingang.auth.CurrentUserProvider
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
@@ -21,9 +22,14 @@ class MyFirebaseMessagingService: FirebaseMessagingService() {
         remoteMessage.data.let {
             val userId = remoteMessage.data["userId"]
             val deeplink = remoteMessage.data["deeplink"]
+            val title = when (remoteMessage.data["notificationType"]) {
+                "new_item" -> applicationContext.getString(R.string.new_item_notification_message)
+                "chat" -> applicationContext.getString(R.string.chat_notification_message)
+                else -> ""
+            }
             if (userId != currentUserProvider.getCurrentUserId() && deeplink != null) {
                 remoteMessage.notification?.let {
-                    sendNotification(it.body!!, deeplink)
+                    sendNotification(it.body!!, title, deeplink)
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


functions/index.js
```diff
@@ -61,8 +71,33 @@ exports.newItemNotification = functions.region(region).firestore.document('items
...
+/**
+ * Gets called every time a chat is added. It obtains the username of the sender from the database, 
+ * then sends a message with the author and the content of the message.
+ * @param change The data that has been updated/created
+ * @param context The context of the updated/created data, contains also the variables (wildcards) used in the document path.
+ * @returns true
+ */
+function onNewChat(change, context) {
+    const newChat = change.data();
+    db.collection("users").doc(newChat.from).get().then((user) => {
```
lumapools, line 86:
```suggestion
    db.collection(DatabaseFields.DBFIELD_USERS).doc(newChat.from).get().then((user) => {
```

lumapools: There are defined constants in DatabaseFields, it's probably better to use them rather than strings (to avoid typos or debugging). I know that here everything is fine but it would just to make sure to use the same thing everywhere 😃

lumapools (changes requested):
Looking good so far! I'm not sure about the "You have received a new message!", I think
`name: message` should be fine. Anyways, let me know what you think. I also requested one small change, but it's looking good. Well done.

functions/index.js
v1ruso: I don't think I can access them from this javascript file

app/src/main/java/com/example/sharingang/utils/notification/MyFirebaseMessagingService.kt
```diff
@@ -21,9 +23,22 @@ class MyFirebaseMessagingService: FirebaseMessagingService() {
...
+            var channelId = ""
+            val title = when (remoteMessage.data["notificationType"]) {
+                "new_item" -> {
+                    channelId = NotificationFields.NEW_ITEM_CHANNEL_ID
+                    applicationContext.getString(R.string.new_item_notification_message)
+                }
+                "chat" -> {
+                    channelId = NotificationFields.CHAT_CHANNEL_ID
+                    remoteMessage.data["fromName"]+":"
+                }
```
cronokirby, line 35: Is it not possible to use the constants you defined earlier for comparison here?

app/src/main/java/com/example/sharingang/utils/notification/MyFirebaseMessagingService.kt
```diff
@@ -21,9 +23,22 @@ class MyFirebaseMessagingService: FirebaseMessagingService() {
...
+                else -> ""
+            }
+            val to = remoteMessage.data["toId"]
+            if ((userId != null && userId != currentUserProvider.getCurrentUserId() || (to != null && to == currentUserProvider.getCurrentUserId())) && deeplink != null) {
```
cronokirby, line 39: Uh, can we simplify this condition, or extract out some meaningful subconditions?

cronokirby (approved):
Looks good, just some minor points to possibly look at

app/src/main/java/com/example/sharingang/utils/notification/MyFirebaseMessagingService.kt
v1ruso: The problem is that those are used first in functions/index.js file, and I haven't found a way to use constants from another file type... If you know how I'll gladly change it

cronokirby: Ah, right, makes sense

app/src/main/java/com/example/sharingang/utils/notification/MyFirebaseMessagingService.kt
v1ruso: You're right, I'll correct that right away!

lumapools (approved):
Looks good. Well done!

PR #280, opened 2021-05-25
Remove useless image access from detailed view

### LINKED ISSUE #278: Fix camera and gallery when editting an item.

bafavre:
Fixes #278 
The image access in the detailed view was useless but since it was registering in detailed, when navigating to edit, the callbacks would only launch on the detailed view.


lumapools (approved):
Thanks for taking the time finding the bug!

v1ruso (approved):
Hackerman

cronokirby (approved):
Bravo!

PR #281, opened 2021-05-25
Added a purchase history 

Superdoude1999:
Added a purchase history to see previously bought items. It is located in the same place as the Sold items, the fragment now serves to display both bought and sold items.

### DIFF
--- a/app/src/main/java/com/example/sharingang/ui/fragments/DetailedItemFragment.kt
+++ b/app/src/main/java/com/example/sharingang/ui/fragments/DetailedItemFragment.kt
@@ -230,7 +230,6 @@ class DetailedItemFragment : Fragment() {
             lifecycleScope.launch {
                 binding.buyButton.isEnabled = true
                 if (paymentProvider.requestPayment(boughtItem, quantity)) {
-                    itemViewModel.sellItem(boughtItem)
                     val newQuantity = boughtItem.quantity - quantity
                     binding.sellerVisibility = if (newQuantity == 0) View.GONE else View.VISIBLE
                     binding.buyButton.isEnabled = newQuantity != 0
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/example/sharingang/ui/fragments/DetailedItemFragment.kt
```diff
@@ -223,19 +223,19 @@ class DetailedItemFragment : Fragment() {
...
             Toast.makeText(context, getString(R.string.incorrect_quantity), Toast.LENGTH_SHORT).show()
-            binding.buyButton.isEnabled = true
         } else {
             lifecycleScope.launch {
-                if (paymentProvider.requestPayment(args.item, quantity)) {
-                    itemViewModel.sellItem(args.item)
-                    val newQuantity = args.item.quantity - quantity
+                binding.buyButton.isEnabled = true
+                if (paymentProvider.requestPayment(boughtItem, quantity)) {
+                    itemViewModel.sellItem(boughtItem)
```
bafavre, line 233: I think you should only set it to sold after the new quantity has been updated.

app/src/main/java/com/example/sharingang/viewmodels/ItemsViewModel.kt
```diff
@@ -69,9 +76,7 @@ class ItemsViewModel @Inject constructor(
     val userItemsAndRequests: LiveData<List<Item>>
         get() = _userItemsAndRequests
 
-    private val _userSoldItems = MutableLiveData<List<Item>>()
-    val userSoldItems: LiveData<List<Item>>
-        get() = _userSoldItems
+//
```
bafavre, line 79: I think you can remove this comment here.

bafavre (approved):
Nice, well done! Just a few remarks :)

app/src/main/java/com/example/sharingang/viewmodels/ItemsViewModel.kt
Superdoude1999: Thanks I removed it

app/src/main/java/com/example/sharingang/ui/fragments/DetailedItemFragment.kt
Superdoude1999: Yeah turns out I didn't need that function in the first place so I just removed it, it was from old code.

lumapools (approved):
Well done, looking good! I'm happy you used the custom radio buttons! Also thank you for providing pictures ❤️

PR #283, opened 2021-05-27
Add weather view

### LINKED ISSUE #282: ## Weather Preview
As a Buyer, I want to see the weather status at a location, so I can avoid bad weather when I go to pickup an item

cronokirby:
image

Fixes #282 

You can now lookup the weather at the location where an item is


app/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,9 @@
         <meta-data
             android:name="com.google.android.geo.API_KEY"
             android:value="@string/google_api_key" />
+        <meta-data
+            android:name="WEATHER_API_KEY"
+            android:value="9ea12af4e349476350e1e525049f75a8" />
```
bafavre, line 23: Will the key be hidden? I don't remember where we're supposed to pu keys.

app/src/main/java/com/example/sharingang/ui/fragments/WeatherFragment.kt
```diff
@@ -0,0 +1,72 @@
...
+            Weather.Condition.Smoke -> R.mipmap.smoke
+            Weather.Condition.Haze -> R.mipmap.haze
+            Weather.Condition.Dust -> R.mipmap.dust
+            Weather.Condition.Fog -> R.mipmap.mist
+            Weather.Condition.Ash -> R.mipmap.ash
+            Weather.Condition.Squall -> R.mipmap.squall
+            Weather.Condition.Tornado -> R.mipmap.tornado
+            Weather.Condition.Clear -> R.mipmap.clear_skies
+            Weather.Condition.Clouds -> R.mipmap.clouds
+            else -> R.mipmap.dunno
```
bafavre, line 68: Maybe put this in another method so codecliamte doesn't complain.

app/src/main/java/com/example/sharingang/viewmodels/ItemsViewModel.kt
```diff
@@ -9,17 +8,26 @@ import androidx.lifecycle.ViewModel
...
 import javax.inject.Inject
+import kotlin.collections.List
+import kotlin.collections.Map
+import kotlin.collections.asReversed
+import kotlin.collections.filter
+import kotlin.collections.forEach
+import kotlin.collections.listOf
+import kotlin.collections.set
+import kotlin.collections.sortedWith
+import kotlin.collections.toMutableMap
```
bafavre, line 30: Were these imports used?

bafavre (changes requested):
Very nice, well done! I'm not sure about having images in app/src/main though, maybe you forgot those?
Also did we really need 284 files just to show the weather, seems a bit over the top. Especially when you can get an icon/image representing the weather directly from openweathermap, that would be better in my opinion.

app/src/main/java/com/example/sharingang/viewmodels/ItemsViewModel.kt
cronokirby: Uh, woops?

app/src/main/java/com/example/sharingang/ui/fragments/WeatherFragment.kt
cronokirby: Good idea

app/src/main/AndroidManifest.xml
cronokirby: If anyone has any ideas I'm open to suggestion. I noticed that our google maps key was there too, so YOLO

cronokirby:
While there may be "284 files", there's only actually 7 images. It's just that when you import an image, Android Studio creates like a bajillion files for different device sizes and formats, etc.

bafavre (approved):
Good job! :)

PR #285, opened 2021-05-28
Block user

lumapools:
Users can now block other users. The block interface is similar to the Report, but the difference is that the information they provide is only for them (i.e. to check a few months later why they have blocked a particular user). More explanations below the pictures.

User Profile View + Block Fragment View
Webp net-resizeimage Webp net-resizeimage (1)

Blocked Users View
Webp net-resizeimage (4) Webp net-resizeimage (3)

Detailed Item View (the item seller has blocked the current user)
Webp net-resizeimage (5)

1. Users cannot block themselves, logged-out users cannot block a user.
    The block consequences are the following: Blocked users won't appear in chats (as there is no point in sending messages to 
    a user one has blocked). Users cannot buy or view the profile of users who blocked them (but the user who has blocked a 
    particular user can still view their profile). 

2. On the Detailed Item View, blocked users don't have a "Buy" button, and the button that directs them to the seller's user profile 
    is unclickable (hence the gray color).

3. On the Blocked Users view, a user can see a list of all users they have blocked. From here, they can either see their profile,
     see the information they provided when they blocked that user (i.e. if they don't remember anymore why they blocked a
     particular user) or unblock that user. A user has no access to a list of users who blocked them.

4. If user A has blocked user B, then none of them will have an entry for the other in ChatsFragment (but the messages are still
    kept).

5. Users can unblock other users.


cronokirby (approved):
There's a lot of code, but it's all pretty straightforward;
excellent work!

bafavre (approved):
:+1:

PR #288, opened 2021-06-01
Use Firebase storage emulator and add project setup instructions

betrisey:


lumapools (approved):
Thanks!

storage.rules
```diff
@@ -0,0 +1,8 @@
+rules_version = '2';
+service firebase.storage {
+  match /b/{bucket}/o {
+    match /user/{uid}/{imageName} {
+      allow read, write: if request.auth!=null && request.auth.uid == uid;
+    }
```
v1ruso, line 6: Shouldn't we add the rules that were already made for cloud.firestore?

betrisey: The users are uploading the images only in /user/{uid}/imageName.jpg so I think that's enough

PR #289, opened 2021-06-03
Fix detection if image was already uploaded

betrisey:
@v1ruso found this bug
When editing an item with an existing image, it doesn't detect that the image was already uploaded when using Firebase Storage emulator.


v1ruso (approved):
Quick and effective, nice!

PR #292, opened 2021-06-03
Fix offline mode

betrisey:


cronokirby (approved):
She sells seashells by the seashore,
The shells she sells are seashells, I’m sure.
So if she sells seashells on the seashore,
Then I’m sure she sells seashore shells.

lumapools (approved):
How much wood would a woodchuck chuck if a woodchuck could chuck wood?

PR #297, opened 2021-06-04
Only allow one click on buy button and add sample data

betrisey:
Sorry you can no longer open 100 buy popups


cronokirby (approved):
Makes sense

v1ruso (approved):
Well done

PR #298, opened 2021-06-04
Fix Firebase sample data on macOS/Linux

betrisey:
The file `sample-data/firestore_export/firestore_export.overall_export_metadata` contained a path with backslashes so it didn't work on Linux and macOS. Now with the slashes, it works on all platforms.


v1ruso (approved):
Nice

cronokirby (approved):
Okie Dokie

PR #299, opened 2021-06-07
Fix offline mode

betrisey:


Superdoude1999 (approved):
Hope that fixes the offline mode but looks correct

v1ruso (approved):
Thanks for your work:)

